PR #48, opened 2020-08-19
Briansbranch

Brian-Jamison:
These are changes needed for windspeed plots (850mb, 250mb, 10m, max updraft (mup), and max downdraft (mdn)).


adb_graphics/default_specs.yml
```diff
@@ -873,3 +886,44 @@ vvort: # Vertical vorticity
...
+wspeed: # Wind Speed
+  10m: &wspeed
+    clevs: !!python/object/apply:numpy.arange [5, 95, 5]
+    cmap: NWS8bitVel
+    colors: wind_colors
+    ncl_name: UGRD_P0_L103_GLC0
+    ticks: 5
+    title: 10m Wind
+#    transform: [conversions.ms_to_kt, windspeed]
+    transform: [vector_magnitude, conversions.ms_to_kt]
```
Brian-Jamison, line 898: This change allows the grib test to just have the 'self' argument.

adb_graphics/default_specs.yml
```diff
@@ -744,10 +744,17 @@ u:
...
+  250mb: &ua_uwind
     ncl_name:
       prs: UGRD_P0_L100_GLC0
       nat: UGRD_P0_L105_GLC0
+    transform: conversions.ms_to_kt
+#  500mb: &ua_uwind
+#    ncl_name:
+#      prs: VGRD_P0_L100_GLC0
+#      nat: VGRD_P0_L105_GLC0
+#    transform: conversions.ms_to_kt
```
christinaholtNOAA, line 756: Please remove all commented code.

adb_graphics/specs.py
```diff
@@ -197,7 +197,24 @@ def lcl_colors(self) -> np.ndarray:
         ctable = ctables.colortables.get_colortable(self.vspec.get('cmap')) \
                     (range(60, 180, 6)) # rainbow
         return ctable
+    @property
```
christinaholtNOAA, line 200: Please add white space between each new function above the `@property` decorator.

adb_graphics/specs.py
```diff
@@ -197,7 +197,24 @@ def lcl_colors(self) -> np.ndarray:
...
+    @lru_cache()
+    def mdn_colors(self) -> np.ndarray:
+
+        ''' Default color map for Max Downdraft '''
 
+        grays = cm.get_cmap('Greys', 2)([0])
+        jet18 = cm.get_cmap('jet', 18)(range(18, 1, -1), alpha=0.6)
```
christinaholtNOAA, line 207: The colormap should be set in the default_specs.yml file and read in here. See lifted_index_colors below for an example.

When choosing names for variables, please choose them so that they are generic and do not hint at the contents of one instance of the variable. The variable `ctable` is used in many of the other configurable color tables described here.

adb_graphics/default_specs.yml
christinaholtNOAA: This seems like a nice direct way to do this. If we aren't going to use the windspeed method, though, maybe we should remove it. Do you mind doing that here since you're handling the windspeed graphics?

Brian-Jamison:
Here are some examples with their NCL counterparts.

image
image
image
image
image
image
image
image
image
image
image
image

PR #50, opened 2020-08-28
add three fields with {fhr} in ncl_name

cshartsough:
Added acpcp, acfrzr, and weasd.  All were fields with {fhr} in the ncl_name.

Plots to follow...

### DIFF
--- a/adb_graphics/default_specs.yml
+++ b/adb_graphics/default_specs.yml
@@ -76,7 +76,7 @@
     unit: dBZ
     vertical_level_name: lv_HTGL7
 acpcp: # Accumulated precipitation
-  sfc: # Not yet supported
+  sfc:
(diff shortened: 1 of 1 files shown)


cshartsough:
Plots.  snow and snow water are empty but color tables are set.  we should think about the height contours on the precip plot.

[Web graphics - PR 50.pptx](https://github.com/NOAA-GSD/adb_graphics/files/5144170/Web.graphics.-.PR.50.pptx)

Brian-Jamison:
Good job on these.  We definitely don't want height contours, but rather MSLP and thickness contours on the precip plots.
Other than that, everything looks great!

christinaholtNOAA:
Looks good!

Brian-Jamison:
Technically, it should read, "remove MSLP overlay from total precip"

Looks good!

PR #53, opened 2020-09-08
Adding the tiles (sub-domains) for maps.

christinaholtNOAA:
Each subdomain is now added to the driver to be done in parallel, along with the forecast hours and different fields.

I did not use exact values of corners from HRRR, instead used less precise estimates.

See the attached power point for examples. We should discuss how to reduce repetition. Also, I'm not super happy with the results when a proper coastline is not present (e.g., CHI-DET).
[DomainCompare.pptx](https://github.com/NOAA-GSD/adb_graphics/files/5190237/DomainCompare.pptx)


Brian-Jamison:
This looks very good!  I'm a bit worried that the current subdomain corners weren't more closely matched, though if it's a problem, these values are easy to find and change.  I'm also concerned with the lack of Lake Michigan coastline in the CHI-DET plot.  It looks like it might be trying to use a dashed line.  Perhaps this is a setting that can be changed in basemap?

cshartsough:
Interesting that the Great Lakes coastlines are plotted for the GreatLakes region map, but not for the CHI-DET and NC region maps. (Also has an issue with Washington state coastlines in the Puget Sound region.)

Is this due to the Great Lakes being along the region boundary, so somehow drawcoastlines() doesn't perform properly? (The dashed line is the actual state boundary running through the middle of the lakes.)  The try block is passing any exception to drawcoastlines(), right?

christinaholtNOAA:
@Brian-Jamison Can you think of any cases where the exact corners of the domains might matter? Are folks just looking at these plots, or are they somehow processing them as a grid?

**About the coastlines and lakes:** 
So, the coastlines code tries to add a LineCollection to the map before it does anything else. For maps that don't have any actual coasts, that collection is empty, it raises a ValueError exception that the try block allows us ignore for the interior domains. Unfortunately, the lake border portion happens in the drawcoastlines() routine AFTER the error. This is the case for North Central, CHI-DET,

I extended the ATL domain to include a bit of coast so that it won't be a problem for that one.

I can get the lakes' outlines if I make all the counties black outlined so that we can take advantage of the fact that the edges of the lakes are also the edges of the counties. Maybe that is the way to go here? Please let me know if this is an acceptable alternative.

cshartsough:
don't see any easy way to trick the code at this point.  Seems like the dark county lines might be the best way  to go for now.

Brian-Jamison:
@christinaholtNOAA The corners that you have are probably ok.  I think the only one that I set via specifications is the VORTEX domain.  There have been a number of occasions in the past where someone has asked me to extend or shift a subdomain in order to capture a particular city or area.  The way you have it set up is easy to change if we get those requests.

On the coastlines, I say try a darker color for the counties.  I think that may work ok.  You can also play with the line thicknesses if that can be done.  I found this https://matplotlib.org/basemap/api/basemap_api.html#mpl_toolkits.basemap.Basemap
and maybe you can set "area_thresh" to get the coastline.

christinaholtNOAA:
My example images didn't upload. :-/


image

image


Thanks @Brian-Jamison. 

I have definitely been playing around with the various different arguments. I played around with the county colors, and since counties don't line up perfectly with the coastlines, it gets a bit disorienting to have the county lines darker. For example:

image
versus the original in the PPT.

image

cshartsough (approved):
looks good.

Brian-Jamison (approved):
I agree that the darker county lines are distracting.  I'd say pick whichever you think works best.
Nice work!

PR #55, opened 2020-09-18
Add Aviation Flight Rules plot

cshartsough:
This is for one new plot, Aviation Flight Rules (flru_sfc).

Added method for aviation_flight_rules
Added special case for text labels on colorbar
Added test in test_grib.py

pylint and pytest both clean

### DIFF
--- a/adb_graphics/datahandler/grib.py
+++ b/adb_graphics/datahandler/grib.py
@@ -292,6 +292,25 @@ class fieldData(UPPData):
         self.level = level
         self.contour_kwargs = kwargs.get('contour_kwargs', {})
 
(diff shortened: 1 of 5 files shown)


cshartsough:
This is for one new plot, Aviation Flight Rules (flru_sfc).

Added method for aviation_flight_rules
Added special case for text labels on colorbar
Added test in test_grib.py

pylint and pytest both clean

Comparison plots:

Brian-Jamison (approved):
Looks great!  Nice job!

christinaholtNOAA (approved):
Code looks great!

If you don't mind, could you move your description in the PR from a "comment" box to the "description" box. I think that first box is the part that gets included with the commit history.

PR #56, opened 2020-09-21
Support for RAP domains.

christinaholtNOAA:
This adds the necessary changes for creating RAP maps.

- Template the NCL variable names to include grid definition (retrieved from first variable entry of file).
- Support Lambert conformal, stereographic, and rotated lat/lon.
- Define single set of sub domains used by both RAP and HRRR.
- Move add_logo from create_graphics.py to figures maps.py to put it consistently in same place for all domains.
- Define wind stride based on domain.
- Update image_lists/hrrr_subset.yml, and alphabetize image lists.

This parallel version will not support timely generation of all graphics in a single job. More work is needed to support that effort.

Passes tests/linter.

### DIFF
--- a/adb_graphics/datahandler/grib.py
+++ b/adb_graphics/datahandler/grib.py
@@ -23,11 +23,12 @@ class GribFile():
 
     ''' Wrappers and helper functions for interfacing with pyNIO.'''
 
-    def __init__(self, filename, filetype):
+    def __init__(self, filename, filetype, **kwargs):
         self.filename = filename
         self.contents = self._load()
 
         self.filetype = filetype
+        self.grid_suffix = kwargs.get('grid_suffix', 'GLC0')
 
     def _load(self):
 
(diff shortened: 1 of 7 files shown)


adb_graphics/datahandler/grib.py
```diff
@@ -138,6 +146,11 @@ def get_level(self, field, level, spec):
 
         ''' Returns the value of the level to for a 3D array '''
 
+        # The index of the requested level
+        lev = spec.get('vertical_index')
+        if lev is not None:
+            return lev
+
```
christinaholtNOAA, line 153: Stop doing work here, if vertical index is set explicitly.

Brian-Jamison:
@christinaholtNOAA why was I removed as a reviewer?

cshartsough (approved):
I could follow most of what you did, so I'm getting better.  ;-)

Didn't see anything off.  A lot of nice work here.

christinaholtNOAA:
@Brian-Jamison I'm not sure. Maybe a random stray click? It wasn't intentional.

cshartsough:
I've had time strange times choosing reviewers, too, having to click and re-click to get it right.  Something odd with GitHub, I guess.

Brian-Jamison:
Ok, I assumed it was unintentional, though Craig is a better reviewer, TBH.
Great work!  I am also understanding more as I go along.

Brian-Jamison (approved):
Awesome sauce!

PR #57, opened 2020-09-22
Windspeed

Brian-Jamison:
These are the specifications for windspeed plots (10m wind, max 10m wind, 80m wind, 850mb wind, 250mb wind, max updraft, and max downdraft).

Examples with their NCL counterparts follow:

wind_10m_f06
wspeed10m_f006
wind_max_f06
wspeedmax_f006
wind_80m_f06
wspeed80m_f006
wind_850_f06
wspeed850mb_f006
wind_250_f06
wspeed250mb_f006
wind_mdn_f06
wspeedmdn_f006
wind_mup_f06
wspeedmup_f006

### DIFF
--- a/adb_graphics/specs.py
+++ b/adb_graphics/specs.py
@@ -190,6 +190,26 @@ class VarSpec(abc.ABC):
 
     @property
     @lru_cache()
+    def mdn_colors(self) -> np.ndarray:
+
+        ''' Default color map for Max Downdraft '''
+
+        grays = cm.get_cmap('Greys', 2)([0])
+        others = cm.get_cmap('jet', 18)(range(18, 1, -1), alpha=0.6)
+        return np.concatenate((others, grays))
+    @property
+    @lru_cache()
+
+    def mup_colors(self) -> np.ndarray:
+
+        ''' Default color map for Max Updraft '''
+
+        grays = cm.get_cmap('Greys', 2)([0])
+        others = cm.get_cmap('jet', 18)(range(1, 18, 1), alpha=0.6)
+        return np.concatenate((grays, others))
+
+    @property
+    @lru_cache()
     def mean_vvel_colors(self) -> np.ndarray:
 
         ''' Default color map for Mean Vertical Velocity '''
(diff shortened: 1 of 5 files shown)


Brian-Jamison:
Sorry, wrong images for 250mb and 850mb.
wind_850_f06
wind_250_f06

adb_graphics/specs.py
```diff
@@ -188,6 +188,26 @@ def lifted_index_colors(self) -> np.ndarray:
...
+
+        ''' Default color map for Max Downdraft '''
+
+        grays = cm.get_cmap('Greys', 2)([0])
+        others = cm.get_cmap('jet', 18)(range(18, 1, -1), alpha=0.6)
+        return np.concatenate((others, grays))
+    @property
+    @lru_cache()
+
+    def mup_colors(self) -> np.ndarray:
```
christinaholtNOAA, line 203: Please rearrange the whitespace here to match the others. This should also be moved down to be in order with the next method.

adb_graphics/specs.py
```diff
@@ -188,6 +188,26 @@ def lifted_index_colors(self) -> np.ndarray:
...
         return ctable
 
+    @property
+    @lru_cache()
+    def mdn_colors(self) -> np.ndarray:
+
+        ''' Default color map for Max Downdraft '''
+
+        grays = cm.get_cmap('Greys', 2)([0])
+        others = cm.get_cmap('jet', 18)(range(18, 1, -1), alpha=0.6)
```
christinaholtNOAA, line 198: The names of the colormaps should be set in default_specs.yml instead of hard coding here. This also applies to the other color definitions below.

christinaholtNOAA (changes requested):
I made a couple of in-line comments on some minor things that should still be addressed below.

Also, you brought up an issue with plotting the correct vector magnitude values in Slack. How was that addressed to get the nice looking plots you attached here?

Brian-Jamison:
Thanks, @christinaholtNOAA, I've fixed the whitespace and cmap issues.

I had realized that the magnitude field is given in the grib2 file explicitly.  It doesn't need to be computed via the components.  The component fields are, however, used in plotting the wind barbs.

cshartsough (approved):
good job, a lot of work here.

Thanks for the thorough review, Christina.

christinaholtNOAA (approved):
Looks good!

cshartsough:
just doing another quick review on GitHub.   Do the u and v max wind fields need an anchor like the other levels?
max:
     ncl_name: MAXUW_P8_L103_{grid}_max1h
     transform: conversions.ms_to_kt
otherwise it looks good.

Brian-Jamison:
No, no anchors necessary, since the magnitude field is available from the grib2 file, we don't have to compute it.

christinaholtNOAA:
Looks good to me!

PR #58, opened 2020-10-09
Adding GitHub Actions workflow

christinaholtNOAA:
My first attempt at adding a GitHub Actions workflow. 

This puts two grib files from HRRR in the repo using git-lfs. The tests are our standard set -- linting and pytest suite. I also updated the environment.yml to match our pygraf environment instead of the older verison of adb_grpaphics. 

I have no idea if this will work!


christinaholtNOAA:
There's a good chance that I need to merge this PR before the actions can be tested.... :(

cshartsough (approved):
Lots in here that's still over my head, but the parts I could see and work through look good.

Brian-Jamison:
I echo Craig's sentiments.

Brian-Jamison (approved):
Looks good!

PR #59, opened 2020-10-14
add precip type plot

cshartsough:
changes for Precip type plot.  Will need further commits to clean up legend code.  pylint expected to crash due to too many local variables.  Sample plots to follow.

### DIFF
--- a/adb_graphics/default_specs.yml
+++ b/adb_graphics/default_specs.yml
@@ -661,22 +661,22 @@ ptyp: # Hourly total precipitation
         linewidths: 0.5
     hatches:
       crain_sfc:
-        colors: ['green']
+        colors: [green]
         hatches: ['', '|||']
         labels: ['Rain']
         levels: [0., 0.1, 1.0]
       csnow_sfc:
-        colors: ['blue']
+        colors: [blue]
         hatches: ['', '---']
         labels: ['Snow']
         levels: [0, 0.9, 1.1]
       cfrzr_sfc:
-        colors: ['purple']
+        colors: [purple]
         hatches: ['', '///']
         labels: ['Freezing Rain']
         levels: [0, 0.5, 1.0]
       cicep_sfc:
-        colors: ['red']
+        colors: [red]
         hatches: ['', '\\\\']
         labels: ['Ice Pellets']
         levels: [0, 0.5, 1.0]
@@ -720,7 +720,6 @@ rh: # Relative Humidity
         linewidths: 0.6
     hatches:
       pres_sfc:
-#        alpha: 0
         hatches: ['...', '']
         levels: [0, 500]
   700mb:
@@ -732,7 +731,6 @@ rh: # Relative Humidity
         linewidths: 0.6
     hatches:
       pres_sfc:
-#        alpha: 0
         hatches: ['...', '']
         levels: [0, 700]
   850mb:
@@ -931,7 +929,6 @@ temp: # Temperature
         colors: grey
     hatches:
       pres_sfc:
-#        alpha: 0
         hatches: ['...', '']
         levels: [0, 500]
     ncl_name:
@@ -953,7 +950,6 @@ temp: # Temperature
         colors: grey
     hatches:
       pres_sfc:
-        alpha: 0
         hatches: ['...', '']
         levels: [0, 700]
     ncl_name:
@@ -973,7 +969,6 @@ temp: # Temperature
         linewidths: 0.6
     hatches:
       pres_sfc:
-#        alpha: 0
         hatches: ['...', '']
         levels: [0, 850]
   925mb:
@@ -986,7 +981,6 @@ temp: # Temperature
         linewidths: 0.6
     hatches:
       pres_sfc:
-#        alpha: 0
         hatches: ['...', '']
         levels: [0, 925]
   sfc:
--- a/adb_graphics/default_specs.yml
+++ b/adb_graphics/default_specs.yml
@@ -665,21 +665,25 @@ ptyp: # Hourly total precipitation
         hatches: ['', '|||']
         labels: [Rain]
         levels: [0, 1]
+        linewidths: 0.1
       csnow_sfc:
         colors: [blue]
         hatches: ['', '---']
         labels: [Snow]
         levels: [0, 1]
+        linewidths: 0.1
       cfrzr_sfc:
         colors: [purple]
         hatches: ['', '///']
         labels: [Freezing Rain]
         levels: [0, 1]
+        linewidths: 0.2
       cicep_sfc:
         colors: [red]
         hatches: ['', '\\\\']
         labels: [Ice Pellets]
         levels: [0, 1]
+        linewidths: 0.2
     ncl_name: APCP_P8_L1_GLC0_acc1h
     ticks: 0
     title: 1 hr Total Precip, Precip Type
--- a/adb_graphics/default_specs.yml
+++ b/adb_graphics/default_specs.yml
@@ -145,7 +145,7 @@ cape:
         linewidths: 0.6
     hatches:
       lpl_agl:
-        alpha: 0
+        alpha: 0.3
         hatches: ['///', '']
         levels: [50, 1000]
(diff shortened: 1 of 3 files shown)


cshartsough:
Pushed a small commit to fix the pylint error and it ran through the tests.  Could still use some changes in default specs, but works and gets through the tests.  Sample plots:
ptyp_sfc_f17
Screen Shot 2020-10-14 at 2 45 15 PM

Brian-Jamison:
Good job!  Lots of work here.  Just curious why linewidths for freezing rain and ice pellets are 0.5 but rain and snow are 0.2?  And does the line collection.set_linewidth(0.1) override this in maps.py?

adb_graphics/default_specs.yml
```diff
@@ -221,6 +221,12 @@ ceilexp2: # Ceiling - experimental no.2
     <<: *ceil
     ncl_name: CEIL_P0_L2_{grid}
     title: Ceiling (exp-2)
+cfrzr: # Categorical Freezing Rain
+  sfc:
+    ncl_name: CFRZR_P0_L1_GLC0
+cicep: # Categorical Ice Pellets
+  sfc:
+    ncl_name: CFRZR_P0_L1_GLC0
```
christinaholtNOAA, line 229: It would be helpful for consistency to add the "grid" template into each of the ncl_name entries you've added here and below.

adb_graphics/default_specs.yml
```diff
@@ -609,6 +621,50 @@ ptmp: # Potential temperature
...
+        hatches: ['', '---', '---']
+        levels: [0, 0.9, 1.1]
+      cfrzr_sfc:
+        colors: ['', 'purple']
+        hatches: ['', '///', '///']
+        levels: [0, 0.5, 1.0]
+      cicep_sfc:
+        colors: ['', 'red']
+        hatches: ['', '\\\\', '\\\\']
+        levels: [0, 0.5, 1.0]
```
christinaholtNOAA, line 662: It might be helpful to arrange these two lists (contours and hatches) in the same order. Now that I'm looking more closely at the code, too, I think we should turn them into actual lists of dicts. Do that by adding a "-" in front of the next heading level under hatches and contours. That _should_ preserve the order of the contours/hatches.

Would you also explore reducing these list to 2 elements? Perhaps the cutoff value and one larger? So, if it's binary, maybe:

```
- cicep_sfc:
  colors: red
  hatches: ['', '\\\\']
  levels: [0.9, 1.1]
```

If this is what it has to be...that's cool, too. This is mostly just my curiosity.

adb_graphics/figures/maps.py
```diff
@@ -301,11 +302,31 @@ def draw(self, show=False):
...
+            if self.field.short_name == 'ptyp':
+                plt.legend(handles=[mpatches.Patch(edgecolor='blue', facecolor='lightgrey', \
+                                                   label='Snow', hatch='----'), \
+                                    mpatches.Patch(edgecolor='red', facecolor='lightgrey', \
+                                                   label='Freezing Rain', hatch='\\\\\\\\'), \
+                                    mpatches.Patch(edgecolor='green', facecolor='lightgrey', \
+                                                   label='Rain', hatch='|||'), \
+                                    mpatches.Patch(edgecolor='purple', facecolor='lightgrey', \
+                                                   label='Ice Pellets', hatch='////')], \
+                                    loc=[0.25, 0.03])
```
christinaholtNOAA, line 329: There's quite a bit of duplicate information between here and the default_specs.yml file. 

You could handle the patches in a list as you loop over hatch_fields like this:
```
colors = field.contour_kwargs.pop('colors')
label = field.contour_kwargs.pop('label')
...
handles.append(mpatches.Patch(edgecolor=colors, facecolor='lightgrey', label=label, 
                             hatch=field.contour_kwargs.get('hatches')[-1])
...
if self.field.short_name == 'ptyp':
    plt.legend(handles=handles, loc=[0.25, 0.03])


```

You'd need to make an empty list for handles outside that loop: 

```
handles = []
```

christinaholtNOAA (commented):
The graphics look great. Excellent job figuring this whole thing out!

I have just a couple of things here that might help us out in the future. I am also a little bit concerned about the order of execution on dictionary items. We might not always get the same results.

adb_graphics/figures/maps.py
cshartsough: Thanks for your comments. Quick responses here, we can discuss in depth at the weekly mtg.

- the linewidths are all leftovers from testing, I had trouble figuring out widths of contour outlines vs widths of hatches.  I'll clean those up.

-  I'll add the {grid} templates to the new fields in default_specs

- I had read while looking at this that ordered dictionaries were now the default (starting with v3.7?), but can add the dashes to make sure, and match the hatch and contour orders.

- I was able to get the hatches color to one item, the hatches to two, but couldn't get the levels to work with less than three.  I can look more into why that is, maybe just a magic combination I haven't found yet.

- I had a loop over hatches/patches in my first commit (when we tested the new linter with intentional errors). To prevent the "TooManyLocalVariables" pylint error, I'll probably have to move this into its own method.

Brian-Jamison (approved):
Looks good!

adb_graphics/figures/maps.py
```diff
@@ -247,6 +248,7 @@ def _colorbar(self, cc, ax):
 
     @utils.timer
     def draw(self, show=False):
+    # pylint: disable=too-many-locals
```
christinaholtNOAA, line 251: Would you mind either adding this to the end of line 250 or indenting it to be consistent with the code inside the method? Either way is good.

adb_graphics/default_specs.yml
```diff
@@ -609,6 +621,54 @@ ptmp: # Potential temperature
...
+        colors: purple
+        levels: [0, 1]
+        linewidths: 0.5
+      cicep_sfc:
+        colors: red
+        levels: [0, 1]
+        linewidths: 0.5
+    hatches:
+      crain_sfc:
+        colors: ['green']
```
christinaholtNOAA, line 648: Can the colors be a string? Will we ever need to have multiples here? 

This change would likely need a change in line below...I'll leave a comment there, too.

adb_graphics/figures/maps.py
```diff
@@ -300,12 +302,29 @@ def draw(self, show=False):
...
         # correspond to a full field of contours.
         if self.hatch_fields:
+            handles = []
             for field in self.hatch_fields:
-                self._draw_field(ax=ax,
-                                 field=field,
-                                 func=self.map.m.contourf,
-                                 **field.contour_kwargs,
-                                 )
+                colors = field.contour_kwargs.get('colors', 'k')
```
christinaholtNOAA, line 307: There is a type mismatch in colors here. If you don't find a 'colors' key (list) in the dict, you get back a string by default.

adb_graphics/figures/maps.py
```diff
@@ -300,12 +302,29 @@ def draw(self, show=False):
...
+
+                cf = self._draw_field(ax=ax,
+                                      field=field,
+                                      func=self.map.m.contourf,
+                                      **field.contour_kwargs,
+                                      )
+
+                # For each level, we set the color of its hatch
+                for i, collection in enumerate(cf.collections):
+                    collection.set_edgecolor(colors[i % len(colors)])
```
christinaholtNOAA, line 321: This line is a bit odd. I'm not sure how many collections we are iterating over, but length of colors is 1 (as defined in the `default_specs.yml` entry), so we'd always get `colors[0]`.

If you change the entry in `default_specs.yml` to a string, you could change this line to just:
```
collection.set_edgecolor(colors)
```

That would get a bit more consistent solution between settings.

I will mention that if you set the `default_specs.yml` setting to a string like `green` you'd get something like `1 % 5 = 1` and `colors[1] = 'r'` which will be an unexpected result.

christinaholtNOAA (approved):
I really like this solution, Craig! Thanks for working through it! :)

I have just a couple small things I think would be nice fixes, if you don't mind.

PR #60, opened 2020-11-02
Better parallel performance with XArray

christinaholtNOAA:
This changes the datahandler grib loader to use XArray with a PyNio engine, and for the UPPData class to have a GribFile object, instead of extending a GribFile object. Benefits of this approach:

- Each grib file is opened once per forecast hour for both maps and Skew Ts
- Faster completion on kjet -- 100 maps for 4 regions in about 3 minutes.
- Locking and error handling on zip file to allow multiple copies to run simultaneously for Rocoto-level parallelization.
- No additional support for Python packages needed.

Tested all current images in hrrr_subset.yml and Skew T locations with create_graphics.py.


adb_graphics/datahandler/grib.py
```diff
@@ -624,15 +636,17 @@ def values(self, level=None, name=None, **kwargs):
...
         profile = field[::]
         lev = 0
         if len(profile.shape) == 2:
             profile = profile[x, y]
         elif len(profile.shape) == 3:
             if one_lev:
-                lev = self.get_level(field, level, var_spec)
+                lev = vertical_index
+                if vertical_index is None:
+                    lev = self.get_level(field, level, var_spec)
```
christinaholtNOAA, line 649: This needs the same logic as fieldData method.

create_graphics.py
```diff
@@ -56,11 +64,48 @@ def create_maps(cla, fhr, grib_path, workdir):
...
+            try:
+                with zipfile.ZipFile(zipf, 'a', zipfile.ZIP_DEFLATED) as zfile:
+                    for png_file in png_files:
+                        if os.path.exists(png_file):
+                            zfile.write(png_file, os.path.basename(png_file))
+            except: # pylint: disable=bare-except
+                print(f'Error on writing zip file! {sys.exc_info()[0]}')
+                count += 1
+                if count >= retry:
+                    raise
```
christinaholtNOAA, line 96: If any error is raised when writing the zip file, it will retry 2 more times.

create_graphics.py
```diff
@@ -56,11 +64,48 @@ def create_maps(cla, fhr, grib_path, workdir):
...
+            else:
+                # When zipping is successful, remove png_files
+                for png_file in png_files:
+                    if os.path.exists(png_file):
+                        os.remove(png_file)
```
christinaholtNOAA, line 101: This block executes if `try` block was successful. Only remove the png files then.

create_graphics.py
```diff
@@ -56,11 +64,48 @@ def create_maps(cla, fhr, grib_path, workdir):
...
+            finally:
+                fd.close()
+                if os.path.exists(lock_file):
+                    os.remove(lock_file)
```
christinaholtNOAA, line 105: Do this block no matter the outcome of the `try` block.

create_graphics.py
```diff
@@ -56,11 +64,48 @@ def create_maps(cla, fhr, grib_path, workdir):
...
+            break
```
christinaholtNOAA, line 106: If successful `try` block, break out of `while True` loop.

create_graphics.py
```diff
@@ -434,22 +477,21 @@ def graphics_driver(cla):
...
-                # Directory is empty now -- rmdir is fine.
-                os.rmdir(workdir)
+                png_files = glob.glob(os.path.join(workdir, f'*{fhr:02d}.png'))
+
+                zip_proc = Process(group=None,
+                                   target=create_zip,
+                                   args=(png_files, zipf),
+                                   )
+                zip_proc.start()
+                zip_proc.join()
```
christinaholtNOAA, line 494: This bit runs the zipping in a subprocess so it doesn't hold up the graphics-making processes.

cshartsough (approved):
tried checking by commit, to follow your changes.  Looks great from all I can tell.

Brian-Jamison (approved):
Cool!  Lots of changes.  I'm a bit confused where the function is located for grib.GribFile.  Is this part of xarray or pynio?

adb_graphics/datahandler/grib.py
```diff
@@ -35,48 +35,33 @@ def _load(self):
         ''' Internal method that opens the grib file. Returns a grib message
         iterator. '''
 
-        return Nio.open_file(self.filename, format="grib2") # pylint: disable=c-extension-no-member
+        #return Nio.open_file(self.filename, format="grib2") # pylint: disable=c-extension-no-member
+        return xr.open_dataset(self.filename,
+                               engine='pynio',
+                               lock=False,
+                               backend_kwargs=dict(format="grib2"),
+                               )
```
christinaholtNOAA, line 43: @Brian-Jamison The `grib.GribFile` class here in `adb_graphics/datahandler/grib.py`. It has been reduced to just opening the grib file and is no longer used as a base class for the other classes defined in this file. 

The only change that really happened to this part of the code is the way we load the data. Before we were using `Nio.open_file()` to get a `Nio` object. Now we use `xr.open_dataset()` with the `pynio` engine and load an XArray data structure. That change in object type was what necessitated a bunch of the changes for accessing the data inside. See the mods below that go from `filename` or`contents` to `ds`.

PR #61, opened 2020-11-19
Mean rh

Brian-Jamison:
These are additions to produce the mean 850mb-500mb RH plot.  A new function, field_mean, was introduced in grib.py to compute the mean of any number of fields of the same type, defined in default_specs.py.  Here is an example with its NCL counterpart:

image
image

### DIFF
--- a/adb_graphics/datahandler/grib.py
+++ b/adb_graphics/datahandler/grib.py
@@ -129,10 +129,18 @@ class UPPData(specs.VarSpec):
 
         return values - self.values(name=variable2, level=level2)
     
-#    def field_mean(self, values, variable2, level2, variable3, level3, **kwargs) -> np.ndarray:
     def field_mean(self, values, variable, levels, **kwargs) -> np.ndarray:
 
-<<<<<<< Updated upstream
+        # pylint: disable=unused-argument
+
+        ''' Returns the mean of the values. '''
+
+        sum = values * 0.
+        for level in levels:
+            sum = sum + self.values(name=variable, level=level)
+            
+        return sum / len(levels)
+    
     def get_field(self, ncl_name):
 
         ''' Given an ncl_name, return the NioVariable object. '''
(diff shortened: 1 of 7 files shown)


christinaholtNOAA (changes requested):
It seems something odd happened with the merge from master here. There are a lot of things that are identical or undoing Craig's previous commits. I don't see a merge from master into this branch? How did you incorporate those changes?

Brian-Jamison:
I'm sorry, I must have done this incorrectly.  On my branch, I committed my latest changes, then I switched to master and did a git pull, which brought in Craig's precip type changes.  I switched back to my branch, and looked at differences between my branch and master, and changed my code to match master plus added my changes.  After testing, I did 'git add' for that code and committed and pushed.  How should I correct it?

image_lists/hrrr_subset.yml
```diff
@@ -83,6 +83,7 @@ hourly:
     rh:
       - 2m
       - 850mb
+      - mean
```
christinaholtNOAA, line 86: This should also show up in any of the other image lists that are relevant...rap, hrrr, etc.

adb_graphics/datahandler/grib.py
```diff
@@ -155,6 +167,9 @@ def get_level(self, field, level, spec):
         levs = self.ds[dim_name].values
 
         # Requested level
+        if level == 'mean':
+            lev = 0
+            return lev
```
christinaholtNOAA, line 172: These lines should not be needed. Provide `vertical_index: 0` in the default_specs.yml instead.

cshartsough (approved):
looks good.  It's really just grib.py and default_specs.yml - I assume hrrr.yml and hrrr_subset.yml are not being kept up in the repository?

christinaholtNOAA (changes requested):
This looks pretty good. Thanks for sorting out the merging issues! 

Please take a look at my inline comments below.

image_lists/hrrr_subset.yml
Brian-Jamison: I'll add it in rap.yml and it should be present for most yml files.

adb_graphics/datahandler/grib.py
Brian-Jamison: I did try that and it does work, but I decided to handle it in grib.py rather than having a vertical_index parameter in the default_specs simply because it looks confusing when just looking at the default_specs.

christinaholtNOAA: I think a quick comment in the default_specs.yml will help with confusion, and keep a bunch of level-type logic out of the code and in the config files.

Brian-Jamison:
This is an example of the new title:
image

cshartsough:
That seemed to do the trick.  Good job.

christinaholtNOAA (approved):
Looks great, @Brian-Jamison!!

PR #62, opened 2020-11-19
Adds an image list for RTMA

christinaholtNOAA:
I created a first pass at the list of RTMA fields to be plotted. I tested in a real-time setting in my own workspace for both maps and skewts. 

The fields that are not currently available are commented out in the image list.


cshartsough (approved):
good cleanup.  should work for RTMA.

Brian-Jamison (approved):
Looks good to me.  Thanks for the cleanup.

PR #63, opened 2020-12-30
Fix the print out of values in TD box.

christinaholtNOAA:
Without this change, the full XArray with all metadata is printed in the thermo-dynamic box.


cshartsough (approved):
This is the level I can understand quickly.  ;-)  Looks good!

PR #64, opened 2021-01-20
Adding RRFS compatibility.

christinaholtNOAA:
These changes allow for compatibility with RRFS UPP output. For maps, that means different Grib2 codes for the "entire atmosphere" fields, and for Skew T plots, allowing thermodynamic fields to be "missing".


cshartsough (approved):
These changes look like they will do the job.  :-)

PR #70, opened 2021-03-16
Run total graphics

christinaholtNOAA:
This addition allows for run-total graphics to be generated for the helicity maps. 

There is an additional flag `--all_leads` that has the graphics driver load all lead times up to the current forecast hour being processed, and needs to be used if attempting to plot run_total or similar types of graphics. The script is expected to work fine if you specify this flag, but do not actually plot any of those types of graphics.

During real time runs, the latest files are opened and the data is appended to any previously loaded xarray dataset to reduce IO.


adb_graphics/datahandler/grib.py
```diff
@@ -210,19 +330,24 @@ def get_transform(self, transforms, val):
...
 
     def latlons(self):
 
         ''' Returns the set of latitudes and longitudes '''
 
-        return [self.ds.coords[c].values for c in \
-                list(self.ds.coords)[-2:]]
+        coords = sorted([c for c in list(self.ds.coords) if
+                         any(ele in c for ele in ['lat', 'lon'])])
+        return [self.ds.coords[c].values for c in coords]
```
christinaholtNOAA, line 350: With all the combining of datasets happening now, there was no set order to expect for finding latitude and longitude. Hopefully all types of files will have lat or lon as part of the coordinate name.

adb_graphics/datahandler/grib.py
```diff
@@ -451,6 +576,14 @@ def grid_info(self):
 
         return grid_info
 
+    def run_total(self, values, **kwargs) -> np.ndarray:
+
+        ''' Sums over all the forecast lead times available. '''
+
+        # pylint: disable=unused-argument,no-self-use
```
christinaholtNOAA, line 583: This one had a conflict between the test, which needed the callable function to be a true method, as opposed to the linter suggesting that since we aren't using "self" it could be a static method. I went with the ability to test, and quieted the linter.

adb_graphics/datahandler/grib.py
```diff
@@ -549,17 +682,35 @@ def values(self, level=None, name=None, **kwargs) -> np.ndarray:
...
+                dim_name = vertical_dim[0]
+
+                if vertical_index is None: # No index is provided in kwargs
                     lev = self.get_level(field, level, spec)
-                vals = field[lev, :, :]
-            else:
-                vals = field[:, :, :]
+
+                try:
+                    vals = vals.isel(**{dim_name: lev})
```
christinaholtNOAA, line 703: This is a more pythonic way of handling grabbing a slice of data, especially when we don't know the dimension order or possible dimensions of the datasets.

adb_graphics/default_specs.yml
```diff
@@ -314,6 +314,10 @@ crain: # Categorical Rain
   sfc:
     ncl_name: CRAIN_P0_L1_{grid}
 cref: # Composite reflectivity
+  esbl:
+    <<: *refl
+    ncl_name: REFC_P0_L10_{grid}
+    title: Ensemble Composite Reflectivity
```
christinaholtNOAA, line 320: This does not yet work.

adb_graphics/utils.py
```diff
@@ -34,6 +34,7 @@ def fhr_list(args):
     args = args if isinstance(args, list) else [args]
     arg_len = len(args)
     if arg_len in (2, 3):
+        args[-1] += 1
```
christinaholtNOAA, line 37: Will allow for inclusive forecast hour -- when running on command line `-f 0 6` previously would only plot 0-5 hour graphics.

adb_graphics/datahandler/grib.py
Brian-Jamison: NCL had the capability of generating the lat, lon arrays when reading gridded data.  I would imagine that PyNio has this same capability, and depending on the grid, the arrays were named lat_0, lon_0, gridlat_0, gridlon_0, so I think you're good, here.

christinaholtNOAA: We are no longer using PyNio as the data handler, but all of these are XArrays with naming conventions provided by PyNio. This is the function that replicates that capability so that we can extract latitude and longitude arrays.

cshartsough (approved):
Went through one commit at a time.  While I wouldn't have been able to come up with all this, I was able to (mostly) follow along with the logic.  Quite a load of work in here, congrats.

create_graphics.py
```diff
@@ -108,6 +105,47 @@ def create_zip(png_files, zipf):
...
+            filetype=cla.file_type,
+            )
+    else:
+
+        # Append a single forecast hour to the existing gribfile object.
+        gribfile.coord_dims.get('fcst_hr').append(fhr)
+        gribfile.append(filenames)
+
+    return gribfile
+
```
christinaholtNOAA, line 148: @Brian-Jamison  is suggesting to use `gribfiles` in this function.

christinaholtNOAA (commented):
Comments during meeting

Brian-Jamison (approved):
Great job!  This is a tough task.  Thanks for walking through it in our meeting.

christinaholtNOAA:
Here are a couple of example plots of 0-2 km helicity, both for the past hour, and for run-total.

image
image
image
image

create_graphics.py
```diff
@@ -317,6 +360,12 @@ def parallel_maps(cla, fhr, ds, level, model, spec, variable, workdir,
         short_name=variable,
         )
 
+    try:
+        field.field
+    except errors.GribReadError:
+        print(f'Cannot find grib2 variable for {variable} at {level}. Skipping.')
+        return
+
```
christinaholtNOAA, line 368: I followed up on this in master, and the code there is currently broken when you try to run all the maps for 0-6 hours. I think it would be best to skip fields we cannot find for that reason. As I mentioned in our meeting last week, that could lead to some tough debugging in the future. Are you both okay with that change, @Brian-Jamison and @cshartsough?

cshartsough: Yes, it all looks good to me.  The discussion was helpful.

Brian-Jamison: I agree.  It's the best way to handle this.  Hopefully we can handle the tough debugging.

PR #71, opened 2021-03-17
Changes for real-time RTMA.  Subdirectories are created for each tile…

Brian-Jamison:
…, and the appropriate images are zipped in each one.

### DIFF
--- a/create_graphics.py
+++ b/create_graphics.py
@@ -376,8 +376,9 @@ def parallel_maps(cla, fhr, ds, level, model, spec, variable, workdir,
 
     # Build the output path
     png_suffix = level if level != 'ua' else ''
-    tile_label = f'_{tile}' if tile != 'full' else ''
+    tile_label = f'_{tile}_' # if tile != 'full' else ''
     png_file = f'{variable}{tile_label}{png_suffix}_f{fhr:03d}.png'
+    png_file = png_file.replace("__", "_")
     png_path = os.path.join(workdir, png_file)
 
     print('*' * 120)
@@ -448,13 +449,13 @@ def graphics_driver(cla):
     '''
 
     # Create an empty zip file
-    zipf = []
-    for tile in cla.tiles:
-        if cla.zip_dir:
+    if cla.zip_dir:
+        zipfiles = {}
+        for tile in cla.tiles:
             tile_zip_dir = os.path.join(cla.zip_dir, tile)
             os.makedirs(tile_zip_dir, exist_ok=True)
             tile_zip_file = os.path.join(tile_zip_dir, 'files.zip')
-            zipf.append(tile_zip_file)				
+            zipfiles[tile] = tile_zip_file
 
     fcst_hours = copy.deepcopy(cla.fcst_hour)
 
@@ -494,23 +495,15 @@ def graphics_driver(cla):
                 create_maps(cla, fhr, grib_path, workdir)
 
             # Zip png files and remove the originals in a subprocess
-            i = 0
-            tile_strings = cla.tiles.copy()
-            tile_strings.reverse()
-            zipf_rev = zipf.copy()
-            zipf_rev.reverse()
-            while i < len(zipf_rev):
-                if tile_strings[i] == "full":
-                    tile_strings[i] = ""
-                png_files = glob.glob(os.path.join(workdir, f'*{tile_strings[i]}*{fhr:02d}.png'))
-
+            for tile, zipf in zipfiles.items():
+                png_files = glob.glob(os.path.join(workdir, f'*_{tile}_*{fhr:02d}.png'))
+                png_files = [png_file.replace("__", "_") for png_file in png_files]
                 zip_proc = Process(group=None,
                                    target=create_zip,
-                                   args=(png_files, zipf_rev[i]),
+                                   args=(png_files, zipf),
                                    )
                 zip_proc.start()
                 zip_proc.join()
-                i = i + 1
 
             # Keep track of last time we did something useful
             timer_end = time.time()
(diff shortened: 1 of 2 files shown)


create_graphics.py
```diff
@@ -447,10 +448,13 @@ def graphics_driver(cla):
     '''
 
     # Create an empty zip file
-    zipf = None
-    if cla.zip_dir:
-        os.makedirs(cla.zip_dir, exist_ok=True)
-        zipf = os.path.join(cla.zip_dir, 'files.zip')
+    zipf = []
+    for tile in cla.tiles:
+        if cla.zip_dir:
```
christinaholtNOAA, line 453: Lines 452 and 453 should be in reverse order for efficiency purposes. Only go through the loop if zip_dir is defined.

create_graphics.py
```diff
@@ -447,10 +448,13 @@ def graphics_driver(cla):
...
+            tile_zip_dir = os.path.join(cla.zip_dir, tile)
+            os.makedirs(tile_zip_dir, exist_ok=True)
+            tile_zip_file = os.path.join(tile_zip_dir, 'files.zip')
+            zipf.append(tile_zip_file)
```
christinaholtNOAA, line 457: The use of a dict for `zipf` would be more effective so that we'd know which tile each was associated with. Also naming it something plural is a bit more clear. Example:

```
zipfiles = {}
for tile in cla.tiles:
    ...
    zipfiles[tile] = tile_zip_file
```

create_graphics.py
```diff
@@ -490,15 +494,23 @@ def graphics_driver(cla):
...
 
             # Zip png files and remove the originals in a subprocess
-            if zipf:
-                png_files = glob.glob(os.path.join(workdir, f'*{fhr:02d}.png'))
+            i = 0
+            tile_strings = cla.tiles.copy()
+            tile_strings.reverse()
+            zipf_rev = zipf.copy()
+            zipf_rev.reverse()
+            while i < len(zipf_rev):
```
christinaholtNOAA, line 502: With the use of a dict as suggested above, this section simplifies to:
```
for tile, zipf in zipfiles.items():
    png_files = glob.glob(os.path.join(workdir, f'*{tile}*{fhr:02d}.png'))
    zip_proc = Process(group=None,
                                   target=create_zip,
                                   args=(png_files, zipf),
                                   )
```

christinaholtNOAA (changes requested):
Thanks for making this work! :)

Looks pretty good, I just have some suggestions for making it a bit more efficient and succinct.

cshartsough (approved):
I like the changes and Christina's suggestions.  Will approve and let you work in the new pieces.  Good job. :-)

create_graphics.py
```diff
@@ -85,6 +85,7 @@ def create_zip(png_files, zipf):
             fd = open(lock_file, 'w')
             print(f'Writing to zip file for files like: {png_files[0][-10:]}')
 
+            print(f'trying zip file {zipf}')
```
christinaholtNOAA, line 88: Would you mind combining this line with the one above to maybe just say "Writing to {zipf} for files like:..."?

create_graphics.py
```diff
@@ -375,8 +376,9 @@ def parallel_maps(cla, fhr, ds, level, model, spec, variable, workdir,
 
     # Build the output path
     png_suffix = level if level != 'ua' else ''
-    tile_label = f'_{tile}' if tile != 'full' else ''
+    tile_label = f'_{tile}_' # if tile != 'full' else ''
```
christinaholtNOAA, line 379: I don't think we need this line anymore. Consider just using "_{tile}_" in line 381 directly.

create_graphics.py
```diff
@@ -447,10 +449,13 @@ def graphics_driver(cla):
...
-    zipf = None
     if cla.zip_dir:
-        os.makedirs(cla.zip_dir, exist_ok=True)
-        zipf = os.path.join(cla.zip_dir, 'files.zip')
+        zipfiles = {}
+        for tile in cla.tiles:
+            tile_zip_dir = os.path.join(cla.zip_dir, tile)
+            os.makedirs(tile_zip_dir, exist_ok=True)
+            tile_zip_file = os.path.join(tile_zip_dir, 'files.zip')
+            zipfiles[tile] = tile_zip_file
```
christinaholtNOAA, line 458: Consider combining lines 457 and 458 for efficiency and brevity.

create_graphics.py
```diff
@@ -490,9 +495,9 @@ def graphics_driver(cla):
                 create_maps(cla, fhr, grib_path, workdir)
 
             # Zip png files and remove the originals in a subprocess
-            if zipf:
-                png_files = glob.glob(os.path.join(workdir, f'*{fhr:02d}.png'))
-
+            for tile, zipf in zipfiles.items():
+                png_files = glob.glob(os.path.join(workdir, f'*_{tile}_*{fhr:02d}.png'))
+                png_files = [png_file.replace("__", "_") for png_file in png_files]
```
christinaholtNOAA, line 500: I'm a little unclear on what this replace is doing. How did we get two underscores in the filename?

christinaholtNOAA (changes requested):

create_graphics.py
Brian-Jamison: Done.  Thanks for spotting that.

create_graphics.py
Brian-Jamison: Thanks!  I don't know why I didn't see that.

create_graphics.py
Brian-Jamison: It was leftover from testing.  The double underscores are taken care of earlier.  That line has been removed.

christinaholtNOAA (approved):

PR #73, opened 2021-04-09
add model_name to maps

cshartsough:
This is a first try to add model names to the maps plots.  It works, but I had to turn off a pylint warning.  Also needs tests for other models besides HRRR, and perhaps handling for no model_name given.

Suggestions welcome.


christinaholtNOAA (approved):

christinaholtNOAA (approved):

Brian-Jamison (approved):
Looks good!

PR #74, opened 2021-04-09
Fix for plotting without zip files.

christinaholtNOAA:
zipfiles is not defined when you don't provide a `-z` command line argument. Put this chunk of code inside an if statement to check for that.


Brian-Jamison (approved):
Ok, this is fine.  I'll have to re-work my fix around this, but it shouldn't be too hard.

cshartsough (approved):
Good idea, I was getting errors on my tests but had ignored them, glad there's a fix now,

PR #75, opened 2021-04-12
Maps model name

cshartsough:
I added the model_name logic to the maps (already done for skewTs), and changed the default to "Unnamed Experiment".

This change also adds shaded hydrometeor regions and filled markers for supercooled layers.  

Passed pylint and pytest, with one pylint check disabled.

Sample skewT:
Screen Shot 2021-04-11 at 7 24 00 PM

### DIFF
--- a/adb_graphics/figures/skewt.py
+++ b/adb_graphics/figures/skewt.py
@@ -56,7 +56,8 @@ class SkewTDiagram(grib.profileData):
         self.max_plev = kwargs.get('max_plev', 0)
         self.model_name = kwargs.get('model_name', 'Analysis')
 
-    def _add_hydrometeors(self, hydro_subplot): # pylint: disable=too-many-locals
+    def _add_hydrometeors(self, hydro_subplot):
+    # pylint: disable=too-many-locals, consider-using-enumerate
         mixing_ratios = OrderedDict({
             'clwmr': {
                 'color': 'blue',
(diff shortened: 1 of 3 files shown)


adb_graphics/figures/skewt.py
```diff
@@ -130,6 +133,26 @@ def _add_hydrometeors(self, hydro_subplot): # pylint: disable=too-many-locals
...
+            if mixr in ['clwmr', 'rwmr']:
+                hydro_subplot.plot(profile[temp.magnitude < 32.0], pres[temp.magnitude < 32.0],
+                                   settings.get('color'),
+                                   fillstyle='full',
+                                   linewidth=0.5,
+                                   marker=settings.get('marker'),
+                                   markersize=6,
+                                  )
+                layer = False
+                for i in range(len(profile)):
```
christinaholtNOAA, line 145: Using enumerate here cleans things up a bit:
```
for i, vertical_lev in enumerate(profile):
    if ((vertical_lev > 0.0 and temp[i].magnitude < 32.0) and not layer):
    ...
```

christinaholtNOAA (commented):
The code looks great! Thanks for tackling the request so quickly.

I am a bit confused about the bounds and the logic, though. The example figure seems to show the profile temperature of 0 C to -18 C (32 F - 0 F) 700 hPa - 500 hPa. Am I missing something there? Maybe the example was a previous iteration highlighting -10 to -20 C?

Brian-Jamison (approved):
Nice job!

adb_graphics/figures/skewt.py
cshartsough: Thanks for showing the proper syntax, I hadn't gotten it right last night.  I've made the change and "un-disabled" the pylint check.

PR #76, opened 2021-04-12
zip_fix

Brian-Jamison:
Recent changes to write images to subdirectories for maps caused a problem when zipping images for skewts.  This is a correction for that problem.

### DIFF
--- a/create_graphics.py
+++ b/create_graphics.py
@@ -506,11 +506,17 @@ def graphics_driver(cla):
 
     # Create an empty zip file
     if cla.zip_dir:
-        zipfiles = {}
-        for tile in cla.tiles:
-            tile_zip_dir = os.path.join(cla.zip_dir, tile)
-            os.makedirs(tile_zip_dir, exist_ok=True)
-            zipfiles[tile] = os.path.join(tile_zip_dir, 'files.zip')
+        if cla.graphic_type == 'skewts':
+            zipf = None
+            os.makedirs(cla.zip_dir, exist_ok=True)
+            zipf = os.path.join(cla.zip_dir, 'files.zip')
+            os.makedirs(cla.zip_dir, exist_ok=True)
+        else:
+            zipfiles = {}
+            for tile in cla.tiles:
+                tile_zip_dir = os.path.join(cla.zip_dir, tile)
+                os.makedirs(tile_zip_dir, exist_ok=True)
+                zipfiles[tile] = os.path.join(tile_zip_dir, 'files.zip')
 
     fcst_hours = copy.deepcopy(cla.fcst_hour)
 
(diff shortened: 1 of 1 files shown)


create_graphics.py
```diff
@@ -546,26 +552,36 @@ def graphics_driver(cla):
...
-            if cla.graphic_type == 'skewts':
-                create_skewt(cla, fhr, grib_path, workdir)
-            else:
-                gribfiles = gather_gribfiles(cla, fhr, gribfiles)
-                create_maps(cla,
-                            fhr=fhr,
-                            gribfiles=gribfiles,
-                            workdir=workdir,
-                            )
-
```
christinaholtNOAA, line 558: This section needs to stay as-is so that the graphics are created no matter whether you intend to zip the files.

create_graphics.py
```diff
@@ -506,11 +506,17 @@ def graphics_driver(cla):
...
     # Create an empty zip file
     if cla.zip_dir:
-        zipfiles = {}
-        for tile in cla.tiles:
-            tile_zip_dir = os.path.join(cla.zip_dir, tile)
-            os.makedirs(tile_zip_dir, exist_ok=True)
-            zipfiles[tile] = os.path.join(tile_zip_dir, 'files.zip')
+        if cla.graphic_type == 'skewts':
+            zipf = None
+            os.makedirs(cla.zip_dir, exist_ok=True)
```
christinaholtNOAA, line 511: These two lines (510 and 511) are unnecessary.

christinaholtNOAA: I think one solution could be to consider a "skewt" a tile. Then you could use the previous chunk of code like this:

```
    if cla.zip_dir:
        zipfiles = {}
        tiles = cla.tiles if cla.graphic_type == "maps" else ['skewt']
        for tile in tiles:
            tile_zip_dir = os.path.join(cla.zip_dir, tile)
            os.makedirs(tile_zip_dir, exist_ok=True)
            zipfiles[tile] = os.path.join(tile_zip_dir, 'files.zip')
```

So that means you'd end up putting your SkewT zip files in `&DATAROOT;/@Y@m@d@H/nclprd/skewt/files.zip` if I am following the RTMA_3D.xml file correctly.

create_graphics.py
```diff
@@ -546,26 +552,36 @@ def graphics_driver(cla):
...
-                create_maps(cla,
-                            fhr=fhr,
-                            gribfiles=gribfiles,
-                            workdir=workdir,
-                            )
-
-            # Zip png files and remove the originals in a subprocess
             if cla.zip_dir:
-                for tile, zipf in zipfiles.items():
-                    png_files = glob.glob(os.path.join(workdir, f'*_{tile}_*{fhr:02d}.png'))
```
christinaholtNOAA, line 562: With the change I mentioned above about using a skewt as a tile, you'd want to change this line:

```
png_files = glob.glob(os.path.join(workdir, f'*{tile}_*{fhr:02d}.png'))
```

Since skewt png files are named like this: `skewt_BAO_99999_f06.png`

christinaholtNOAA (commented):
I think your code looks good for getting this to work. I made a couple of suggestions for how to think about it differently and make far fewer changes to what is currently in the main branch.

create_graphics.py
```diff
@@ -557,15 +558,14 @@ def graphics_driver(cla):
...
-                    zip_proc.start()
-                    zip_proc.join()
+            for tile, zipf in zipfiles.items():
+                png_files = glob.glob(os.path.join(workdir, f'*_{tile}_*{fhr:02d}.png'))
+                zip_proc = Process(group=None,
+                                   target=create_zip,
+                                   args=(png_files, zipf),
+                                   )
+                zip_proc.start()
+                zip_proc.join()
```
christinaholtNOAA, line 568: This section needs to be inside an if statement: `if cla.zip_dir` as it was before.

christinaholtNOAA (changes requested):
Brian, this looks pretty good. Just needs one fix that got left out.

create_graphics.py
Brian-Jamison: Thanks, I must have missed that.  I added it back in.

christinaholtNOAA (approved):
Looks great!

cshartsough (approved):
Looks clean.  Nice job.

christinaholtNOAA (approved):
Looks great! Thanks, Brian!

PR #77, opened 2021-04-15
Modifications to make sure all our current models are working.

christinaholtNOAA:
PEP8 standards required splitting the adb_graphics/datahanlder/grib.py module into two as it had gotten too long.

The vertical_index flag is used in far fewer variables now, since we can't expect levels to be consistent across model output files. For example, 06km level now splits the level label into a list and looks for the vertical levels defined in the file that correspond to [0, 6000]

Added functionality to define a list of potential NCL names since hrrr and hrrrx define a few differently. 

Removed a couple of variables from image lists to match what is available in real-time with operational rap/hrrr files.

### DIFF
--- a/adb_graphics/datahandler/gribdata.py
+++ b/adb_graphics/datahandler/gribdata.py
@@ -11,7 +11,6 @@ from string import digits, ascii_letters
 
 from matplotlib import cm
 import numpy as np
-import xarray as xr
 
 from .. import conversions
 from .. import errors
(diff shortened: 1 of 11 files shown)


cshartsough (approved):
You're right, a lot of pieces in there, they all fit in with our discussion on Wednesday,  Looks great.

Brian-Jamison (approved):
Wow, lots of changes!  Looks good!

PR #79, opened 2021-04-27
Previous PR Broke RAP

christinaholtNOAA:
That line is evaluated, but variable doesn't exist for RAP


cshartsough (approved):
thumbs up

christinaholtNOAA:
There were more fixes needed. Opened a new PR that also has these changes.

PR #80, opened 2021-04-30
Change output filenames for Skew-T to use 3 digit forecasts

Brian-Jamison:
This is literally a one-character change in the code.  It makes the skew-t output file names end in "fNNN.png" rather than "fNN.png".


cshartsough (approved):
guess I can trust a one-character change.  fingers crossed.

christinaholtNOAA (approved):

christinaholtNOAA:
@Brian-Jamison I approved, but am curious if this change will cause issues for our other CAM models.

PR #82, opened 2021-05-04
small changes to resize the maps and NOAA logo

cshartsough:
Resets the map size from 12x12 to 10x10, to reduce the size of zip files.

Also use the smaller NOAA logo to better fit the new plot size.

Passed pylint but not pytest, will look into that if it fails on GitHub tests.

Example plots, screenshots so not sure if the resizing will be apparent.  Reduced sizes from roughly 450 kb to 350 kb (80% of original)

Old:
Screen Shot 2021-05-04 at 2 47 06 PM

New:
Screen Shot 2021-05-04 at 2 47 43 PM

### DIFF
--- a/adb_graphics/figures/maps.py
+++ b/adb_graphics/figures/maps.py
@@ -207,7 +207,7 @@ class DataMap():
 
         ''' Puts the NOAA logo at the bottom left of the matplotlib axes. '''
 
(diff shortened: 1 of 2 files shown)


Brian-Jamison (approved):
Thanks for this, Craig!

christinaholtNOAA (approved):
lgtm

PR #83, opened 2021-05-05
Fixes needed for run-total graphics in HRRR

christinaholtNOAA:
A majority of the changes grib data were understanding-related and linting. While I was debugging, I couldn't follow the naming convention I had previously chosen for all the diff levels (requested, ones from the grib file, etc.) so I tried to clarify those a bit.

The real fix is for reading in the data. By defining a limited set of variables to rename and concatenate, we were leaving out all other variables, whose values defaulted to the 0-hr forecast values. Now we will concatenate all the variables in the dataset, and find the "accumulated" variables that should be renamed in the 0- and 1-hr forecast files based on their suffixes.

I tested this with the `--all_leads` flag using two forecast lead times, e.g. `-f 0 6` and it works out, however, the functionality to do a single lead time with the `--all_leads` flag is broken. Running with a single forecast lead time and no `--all_leads` flag works except that the accumulated variables create maps that say they are accumulated over the whole forecast lead time, but are in fact only over the past hour. 

I plan to open issues for these two cases and return to the capabilities later, as they are less important than the real-time capabilities that should be fixed with these mods.


Brian-Jamison (approved):
Nice!  Thanks for making these changes.

cshartsough (approved):

PR #88, opened 2021-05-14
add rtma_ncep.yml to image_lists

Brian-Jamison:
Added a new file to image_lists, rtma_ncep.yml, since products are limited from those files.  Also avoids an error resulting from the 2m - skin temperature plot, since only 2m temperature is available.


christinaholtNOAA (approved):
Looks good!

cshartsough (approved):
nice job. :-)

PR #89, opened 2021-05-17
Rename a few odd accumulated variables in free forecast

christinaholtNOAA:
A few RRFS variables start having a "1h" suffix after 6 hours. I am renaming all of those to remove that for consistency when opening the dataset.

Tested with RRSF_dev1 workflow.


Brian-Jamison (approved):
Looks good, thanks!  It'll be nice to not worry about these naming inconsistencies.

cshartsough (approved):
thought I had submitted my approval earlier. in any case, looks good.

PR #90, opened 2021-05-19
Temp colors quickfix

cshartsough:
This PR is to change the default color table for 2m and sfc Temp, and add annotated values at airports (default is False, only set for 2m T here).

The version failed one pytest step and 2 pylint (unused variable and too-many-variables), the latter was disabled.

Some tweaking of the colors may still be needed, although this is not the final color table so might wait on that step.

Sample 2m T plot:
Screen Shot 2021-05-19 at 4 10 54 PM


Brian-Jamison (approved):
Looks good!  Great job!

christinaholtNOAA (approved):
Looks great! I saw there was one other pylint issue  for an unused variable. Otherwise, seems good to go. :)

PR #91, opened 2021-05-24
change map resolution to reduce zip file size.

cshartsough:
This is a one-word change in create_graphics.py, to reduce the plot sizes and allow for smaller overall zip file size.  

passed both pylint and pytest.

adding sample plots, not sure how they'll look in GitHub (each display software behaves differently).  On the Web pages, expecting the plots to be reduced in size with little loss of resolution at the new size.
Screen Shot 2021-05-24 at 11 09 14 AM
Screen Shot 2021-05-24 at 11 09 26 AM

### DIFF
--- a/create_graphics.py
+++ b/create_graphics.py
@@ -489,7 +489,7 @@ def parallel_skewt(cla, fhr, ds, site, workdir):
     plt.savefig(
         png_path,
         bbox_inches='tight',
-        dpi='figure',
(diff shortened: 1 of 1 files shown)


Brian-Jamison (approved):
Thanks for doing this, looks good!

christinaholtNOAA (approved):
LGTM!

PR #95, opened 2021-06-10
Fixes for accumulated precip

christinaholtNOAA:
This set of changes renames the variables in RAP/HRRR/RRFS that include the forecast in the accumulated variables to only be subscripted with "acc". This allows xarray to accumulate them as a single variable. It should fix the absence of the plots on the web.

I also changed the way longer lists of contoured variables were plotted so that the titles wouldn't as often run long. Here's an example of 1hr total precip (from two different times).

Update:
image

Current:
image


Brian-Jamison (approved):
LGTM!

PR #96, opened 2021-06-18
Fix to check for & remove existing zip files

Brian-Jamison:
When a run needs to be restarted, new files get appended to existing zip files from the previous run, despite having the same names. This fix checks for existing zip files and removes them if present.

### DIFF
--- a/create_graphics.py
+++ b/create_graphics.py
@@ -510,6 +510,8 @@ def graphics_driver(cla):
 
     '''
 
+    # pylint: disable=too-many-branches, too-many-locals
+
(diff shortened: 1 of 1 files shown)


christinaholtNOAA (approved):
LGTM!

Brian-Jamison:
I made some changes for pylint.  I added a line in graphics_driver to disable too-many-locals and too-many-branches.  I only added one variable and one branch.

cshartsough (approved):
seems like a good fix.  congrats.

PR #97, opened 2021-06-29
Fix renaming conditions to be more strict.

christinaholtNOAA:
This fix was applied in real time runs for HRRR_NCEP and seems to be working as expected.


cshartsough (approved):
woo hoo!  ahem, I mean, good job.  Glad we found the issue, hope it's all set now.  Good work figuring out a tough set of clues.

PR #98, opened 2021-07-06
Adding HRRR Smoke Products

christinaholtNOAA:
This batch of graphics adds the list for HRRR Smoke graphics, and sets up the specs for new fields (except for fire weather index). I have already fixed the units on the Near-sfc Smoke plot, but did not regenerate the figures.

I've also modified a few units to use Latex interpretation for nicer display of titles.

Examples of new graphics:

image
image

image
image

image
image

Example of new unit formatting:
image

### DIFF
--- a/adb_graphics/conversions.py
+++ b/adb_graphics/conversions.py
@@ -72,6 +72,12 @@ def percent(field: np.ndarray, **kwargs) -> np.ndarray:
 
     return np.asarray(field) * 100.
 
+def to_micro(field: np.ndarray, **kwargs) -> np.ndarray:
+
+    ''' Convert field to micro '''
+
+    return np.asarray(field) * 1E6
+
 def vvel_scale(field: np.ndarray, **kwargs) -> np.ndarray:
 
     ''' Scale vertical velocity for plotting  '''
--- a/adb_graphics/datahandler/gribdata.py
+++ b/adb_graphics/datahandler/gribdata.py
@@ -526,37 +526,35 @@ class fieldData(UPPData):
 
         '''
 
-        pres_sfc = self.values(name='pres', level='sfc') * 100. # convert back to Pa
-        temp = self.values(name='temp', level='2m') + 273.15   # convert back to K
-        sphum = self.values(name='sphum', level='2m').values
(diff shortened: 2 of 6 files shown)


christinaholtNOAA:
After chatting with @cshartsough, we decided I would merge his work with Fire Weather Index into this PR...working toward that now.

Brian-Jamison:
Looks good!  And good to merge in the fwi plots.  My only complaint is the size of the dots for fire radiative power, which need to be a little bigger, at least for the conus domain.  Also, we can save some plotting time by only plotting fire radiative power for the analysis, because this field doesn't change during the forecast.

christinaholtNOAA:
Okie dokie. Here are the NCL and Python Fire Weather Index (newly named Fire Potential) plots. @EricJames-NOAA said the black contours in the NCL were unnecessary in a Slack convo on the topic.

image

image

cshartsough (approved):
Reviewed this before being added as a reviewer.  ;-)  It looks good.

christinaholtNOAA:
@Brian-Jamison I'm coming back around to your earlier comments, finally.

The dots for the Fire Radiative Power are the continuous field, and how Python renders it. I will take a look at the NCL script again to see if there is anything going on that isn't entirely obvious about how to get them to be more similar. 

We don't currently have the capability to not plot longer-lead time plots. If the field is available, it will be plotted at all the requested lead times.

Brian-Jamison:
The NCL script does a nested loop to look at each grid point and determine what threshold it falls in, then plots a marker (filled circle) of a certain size at that grid point.  It's not elegant, admittedly, but it works.
And no big deal on limiting the FRP plot to just the analysis, I think it's good to have it at all forecast times for convenience.

christinaholtNOAA:
Oh. That's different. We definitely don't have the capability of doing scatter plots over the maps.

Brian-Jamison (approved):
LGTM!

cshartsough:
You may have seen Eric's request to change the name from "Hourly Fire Potential" to "Hourly Wildfire Potential".  I think (??) that's a quick change for this PR, before it gets merged?

christinaholtNOAA:
I did see that. Just pushed the one-liner. When/if the checks pass, I can merge and consider deploying a little later today.

PR #99, opened 2021-07-08
Add subdomains for HRRR-AK

Brian-Jamison:
Added tiles for subdomains in HRRR-AK.  Arguments for --tiles should be 'full AKRange Anchorage Juneau' and not 'all'.


cshartsough (approved):
Seems like the correct method.  I'm thinking the idea is, for AK regions, specify them, and the "all" flag is to get regions excluding the rap and AK only ones.

christinaholtNOAA (approved):

Brian-Jamison:
Yes, I think we'd like to preserve the 'all' flag just for CONUS subdomains.

PR #100, opened 2021-07-14
Ensemble plots

cshartsough:
Initial changes for ensemble plots.  This handles most of the single-grib-file plots, including flash flood and total precip probabilities for various fcst hours and thresholds.  Also, total ensemble precip for various forecast hours.

Sample plots below, the main difference is the color table, and we don't include a max value crosshair and printed value.

Passed pylint and pytest (with changes to test_common.py).

image
image

image
image

image
image


adb_graphics/datahandler/gribfile.py
```diff
@@ -213,3 +213,52 @@ def open_kwargs(self):
             engine='pynio',
             lock=False,
             )
+
```
cshartsough, line 216: not sure why the variable_names method is in here but not in main, I thought I was up to date on pulls.  I can remove it if it shouldn't be there.

Brian-Jamison (approved):
Looks good!

adb_graphics/datahandler/gribfile.py
christinaholtNOAA: Yeah, I think this method should be removed. Maybe check to see if it breaks anything when running a few examples?

adb_graphics/default_specs.yml
```diff
@@ -115,6 +115,111 @@ acpcp: # Accumulated run total precipitation
...
+    clevs: [0., 0.01, 0.1, 0.25, 0.5, 1, 2, 3, 4, 6, 10]
+    cmap: gist_ncar
+    colors: rainbow12_colors
+    ncl_name: APCP_P12_L1_{grid}_acc6h
+    ticks: 0
+    title: Ensemble Mean 6-hr Precipitation
+    transform: conversions.kgm2_to_in
+    unit: in
+acpcpens1: # ensemble probability of precipitation, 1h
+  76.2mm: &ens_precip_prob1
```
christinaholtNOAA, line 146: I'm curious if we could get even more re-use of this anchor here? Could we just use `&ens_precip_prob` and then change only the ncl_name in the different accumulation bins below?

christinaholtNOAA (changes requested):
Sorry it took a while to get to this one. The code looks great, just a couple of places with opportunity for cleanup.

adb_graphics/datahandler/gribfile.py
cshartsough: Seems to be fine without this method. Leaving it out to look like the main branch.

adb_graphics/default_specs.yml
cshartsough: made that change, and also added another anchor for ensemble means.

christinaholtNOAA (approved):
Thanks for making those adjustment! Looks great!

PR #101, opened 2021-07-19
Adjust titles

Brian-Jamison:
These are some changes for the titles.  The biggest changes are that the main plot title is in the top center, and valid time has been moved to the top left near the init time.  Hatched and Contoured labelling appears at the top right, when necessary.  The colorbar label font has been reduced to avoid overlapping.  Also there are a few adjustments for superscripts.

Old:
image

New:
image

Old:
image

New:
image


Brian-Jamison:
Forgot to add the reviewers.

adb_graphics/default_specs.yml
```diff
@@ -1340,11 +1340,11 @@ vvort: # Vertical vorticity
     colors: vort_colors
     ncl_name: RELV_P8_2L103_{grid}_max1h
     ticks: 0
-    title: 0-1km Max Vertical Velocity (over prev hour)
+    title: 0-1km Max Vertical Vorticity (over prev hour)
     unit: 1/s
   mx02: # Hourly maximum of vertical vorticity over 0-2 km layer
     <<: *vvort 
-    title: 0-2km Max Vertical Velocity (over prev hour)
+    title: 0-2km Max Vertical Vorticity (over prev hour)
```
Brian-Jamison, line 1347: I can't believe that we haven't caught this error yet.

adb_graphics/figures/maps.py
```diff
@@ -249,8 +249,10 @@ def _colorbar(self, cc, ax):
 
         if self.field.short_name == 'flru':
             ticks = [label.rjust(30) for label in ['VFR', 'MVFR', 'IFR', 'LIFR']]
+        if self.field.short_name == 'ulwrf' and self.field.level == 'sfc':
```
cshartsough, line 252: not sure if would be useful here, but there is a way to set the number of tick labels in default_specs, I think it's:
- ticks > 0, set every "ticks" units, e.g., ticks=10, for temp, label every 10 deg.
- ticks=0, no tick labels
- ticks<0, label every "ticks" intervals., e.g. ticks=-2, label every other color level.

cshartsough (approved):
Just added one comment about the tick levels, not a big deal.  The rest looks good, as we discussed last week.

Brian-Jamison:
Thanks, @cshartsough, I forgot about that.  I've already tested that and it works.  I'll wait for @christinaholtNOAA to comment, and make any changes she notices.

adb_graphics/default_specs.yml
christinaholtNOAA: Wow! Nice catch!

adb_graphics/figures/maps.py
christinaholtNOAA: I think that would be the best. We don't want multiple ways of doing the same things.

adb_graphics/figures/maps.py
```diff
@@ -406,48 +408,48 @@ def _title(self):
...
         if self.contour_fields:
             for cf in self.contour_fields:
                 if cf.short_name not in not_labeled:
                     title = cf.vspec.get('title', cf.field.long_name)
+                    title = title.replace("Geopotential", "Geop.")
                     contoured.append(f'{title}')
                     contoured_units.append(f'{cf.units}')
 
-        contoured = ', '.join(contoured)
+        contoured = ''.join(contoured)  # The join method makes contoured a string, enabling linefeeds
```
christinaholtNOAA, line 422: Did you check out an example of the title on a plot that has multiple contours and hatches?

adb_graphics/figures/maps.py
```diff
@@ -406,48 +408,48 @@ def _title(self):
...
             units = f''
-        plt.title(f"{title} {units}\n {contoured}",
+
+        # Title or Atmospheric level and unit in the high center
+        if f.vspec.get('title'):
+            title = f.vspec.get('title')
+            plt.title(f"{title} {units}", position=(0.5, 1.06), fontsize=18)
+        else:
+            level = level if not isinstance(level, list) else level[0]
+            plt.title(f"{level} {lev_unit} {f.field.long_name} {units}", position=(0.5, 1.06), fontsize=18)
```
christinaholtNOAA, line 445: To reduce duplication, can you rearrange this so that we only have a single `plt.title(.....` line? If we wanted to move that tile line after this change, we'd need to edit positions in two places, for example.

christinaholtNOAA (changes requested):
These changes look really nice! Just a few questions and some minor changes I requested below.

Brian-Jamison:
A couple of examples after correcting:

image

image

christinaholtNOAA (approved):

PR #103, opened 2021-07-26
Changes for HRRR-HI

Brian-Jamison:
An adjustment to figure size is needed for HRRR-HI in order to keep titles from overlapping.  Also made an adjustment for the wind barb stride since the projection is 'lcc' and domain is 'full'.

image


christinaholtNOAA (approved):
Looks good!

cshartsough (approved):
nice work.   I hope we don't get too many oddball areas over time.

PR #104, opened 2021-07-27
Hrrr hawaii

Brian-Jamison:
I missed precip type and 850mb wind in the hrrrhi_subset.yml so I added them.


cshartsough (commented):
good catch

christinaholtNOAA (approved):

PR #106, opened 2021-08-09
Adding native level smoke plots

Brian-Jamison:
Added changes to default_specs.yml to plot 1000ft AGL and 6000ft AGL smoke with winds.

image
image
image
image

### DIFF
--- a/adb_graphics/default_specs.yml
+++ b/adb_graphics/default_specs.yml
@@ -1301,7 +1301,7 @@ trc1:
     title: Vertically Integrated Smoke
     transform: conversions.to_micro
     unit: $mg/m^2$
-  lvl4: &tracer1
+  1000ft: &tracer1
     clevs: [1, 2, 4, 6, 8, 12, 16, 20, 25, 30, 40, 60, 100, 200]
     colors: smoke_colors
     ncl_name: MASSDEN_P0_L105_{grid}
@@ -1309,9 +1309,11 @@ trc1:
     title: 1000ft AGL Smoke
     wind: True
     unit: $\mu g/m^3$
-  lvl11:
+    vertical_index: 4
+  6000ft:
     <<: *tracer1
     title: 6000ft AGL Smoke
+    vertical_index: 11
   sfc:
     clevs: [1, 2, 4, 6, 8, 12, 16, 20, 25, 30, 40, 60, 100, 200]
     colors: smoke_colors
@@ -1334,12 +1336,14 @@ u:
   700mb: *ua_uwind
   850mb: *ua_uwind
   925mb: *ua_uwind
-  lvl4:
+  1000ft:
     ncl_name: UGRD_P0_L105_{grid}
     transform: conversions.ms_to_kt
-  lvl11:
+    vertical_index: 4
+  6000ft:
     ncl_name: UGRD_P0_L105_{grid}
     transform: conversions.ms_to_kt
+    vertical_index: 11
   max:
     ncl_name: MAXUW_P8_L103_{grid}_max1h
     transform: conversions.ms_to_kt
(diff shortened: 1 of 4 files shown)


adb_graphics/default_specs.yml
```diff
@@ -1326,6 +1344,12 @@ u:
   max:
     ncl_name: MAXUW_P8_L103_{grid}_max1h
     transform: conversions.ms_to_kt
+  lvl5:
```
cshartsough, line 1347: do these winds need to be converted to AGL?  I know we have transforms to AGL for other fields like Aviation Flight Rules, ceiling, etc.  There seem to be difference in the NCL v python winds, especially over the mountains.

adb_graphics/default_specs.yml
```diff
@@ -1301,6 +1301,24 @@ trc1:
...
+    clevs: [1, 2, 4, 6, 8, 12, 16, 20, 25, 30, 40, 60, 100, 200]
+    colors: smoke_colors
+    ncl_name: MASSDEN_P0_L105_{grid}
+    ticks: 0
+    title: 1000ft AGL Smoke
+    vertical_index: 4
+    wind: True
+    unit: $\mu g/m^3$
+  lvl12:
+    clevs: [1, 2, 4, 6, 8, 12, 16, 20, 25, 30, 40, 60, 100, 200]
```
christinaholtNOAA, line 1314: Please consider using an anchor and reference to reduce duplication of information here.

adb_graphics/default_specs.yml
```diff
@@ -1301,6 +1301,24 @@ trc1:
     title: Vertically Integrated Smoke
     transform: conversions.to_micro
     unit: $mg/m^2$
+  lvl5:
```
christinaholtNOAA, line 1304: I think that if you set the level here to `1000ft` and `6000ft` below, then you shouldn't need to provide a vertical_index. I've run into issues providing vertical index explicitly, especially when different models/configs output different sets of levels.

I'm not sure what the value of levels are in the vertical dimension. Maybe it's kilofeet? If needed, a multiplier can be applied inside the method `numeric_level` in the `datahandler.gribdata` package.

christinaholtNOAA (approved):
Nice work. The graphics look good. 

I have just a couple of comments below. I think that we should use the ability to find the level more autonomously where we can so that it's more applicable across various configurations of grib output.

adb_graphics/default_specs.yml
Brian-Jamison: It turns out that these levels for the python plots are one level higher than they should be.  Please see my reply to Christina for the details.

adb_graphics/default_specs.yml
Brian-Jamison: I was struggling with anchors before, and planned to revisit it once I got things straightened out.  The next commit will have that.

adb_graphics/default_specs.yml
Brian-Jamison: So I discovered that I didn't need vertical_index because using the names "lvl5" and "lvl12" were parsed making the levels 5 and 12.  But the actual level indices that I need are 4 and 11.  I changed these to "lvl4" and "lvl11" in the next commit.

The index of the vertical level in the native files doesn't have a unit.  The monikers "1000ft" and "6000ft" are approximations, so I can't provide that as a level.

It looks like the test_common.py gives me an assertion error related to "lvl4" in my latest test.  The graphics run fine, but the pytest generates this error.  Is this something that can be corrected in test_common.py?

Brian-Jamison:
Names were changed from "lvl5" and "lvl12" to "1000ft" and "6000ft".  vertical_index was explicitly provided.  Put in an anchor for the definitions of these in default_specs.yml.  Here are some new graphics to compare:

image
image
image
image

cshartsough:
The wind magnitudes don't seem to match, especially for 1000ft.  smoke patterns are similar but maybe off a bit as well.  Maybe still having an issue with the level selected?

Brian-Jamison:
I believe that maybe some round-off characteristics, along with wind barbs that don't exactly coincide, could be responsible for the differences in winds.  Also, the threshold differences in plotting between NCL and Python could account for the some of the smoke contour differences.  I'll re-run and post graphics from level 3 and 5, just to be sure.

Brian-Jamison:
This is level 3:

image


This is level 5:

image

cshartsough (approved):
code looks good - not sure why the winds look a little different but we'll try to keep an eye on things.

PR #107, opened 2021-08-16
Hrrr car

Brian-Jamison:
Testing for HRRR-CAR showed some title-overrunning with the main title and lower titles.  This effect is also seen in the current HRRR-NCEP full domain plots.  The title offset has been increased to fix this.  Also, testing revealed that titles with more than 45 characters causes the data plotting area to become smaller.  Some abbreviation has been put in to remedy this problem.

### DIFF
--- a/adb_graphics/default_specs.yml
+++ b/adb_graphics/default_specs.yml
@@ -682,42 +682,42 @@ hlcytot:
   mn02:
     <<: *hlcy_mn02
     accumulate: True
-    title: Run Total 0-2km Min Updraft Helicity (over prv hr)
+    title: Run Total 0-2km Min Updraft Helicity
     transform: run_total
   mn03:
     <<: *hlcy_mn03
     accumulate: True
-    title: Run Total 0-6km Min Updraft Helicity (over prv hr)
+    title: Run Total 0-6km Min Updraft Helicity
     transform: run_total
   mn16:
     <<: *hlcy_mn16
     accumulate: True
-    title: Run Total 1-6km Min Updraft Helicity (over prv hr)
+    title: Run Total 1-6km Min Updraft Helicity
     transform: run_total
   mn25:
     <<: *hlcy_mn25
     accumulate: True
-    title: Run Total 2-5km Min Updraft Helicity (over prv hr)
+    title: Run Total 2-5km Min Updraft Helicity
     transform: run_total
   mx02:
     <<: *hlcy_mx02
     accumulate: True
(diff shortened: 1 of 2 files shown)


cshartsough (approved):
Looks good. two small comment/questions:
1. you removed the "(over prv hr)" portion of most titles, but left if for graupel (abbreviated). Would it make sense to remove it there as well, for consistency?
2. do you have  comparison plot to show the better title alignment?  It's minor, so not a big deal.

adb_graphics/default_specs.yml
```diff
@@ -982,7 +982,7 @@ rh: # Relative Humidity
   mean: # Mean RH 850mb to 500mb
     <<: *rh
     print_units: false
-    title: Mean 850-500mb Relative Humidity (%, shaded), 700mb Wind (kt)
+    title: Mean 850-500mb RH (%, shaded), 700mb Wind (kt)
```
christinaholtNOAA, line 985: What does this one do about the units? I was thinking we didn't typically specify units in the title and they were handled by the script.

adb_graphics/default_specs.yml
```diff
@@ -1355,29 +1355,29 @@ ulwrf: # Upward Longwave Radiation Flux
     colors: radiation_colors
     ncl_name: ULWRF_P0_L1_{grid}
     ticks: -2
-    title: Upward Longwave Radiation Flux, Surface
+    title: Upward LW Radiation Flux, Surface
```
christinaholtNOAA, line 1358: I think it would be more clear to abbreviate `surface` to `sfc` than `Longwave` here. Same on the variables below.

christinaholtNOAA (approved):
I pointed out a couple of very minor changes here, but looks good. Thanks!

Brian-Jamison:
Good questions.  I removed "(over prv hr)" only from the Run-Total plots, since the plot is a run total, and (except for the 1hr forecast) is misleading.  That's not the case for graupel, which is not a run total (actually the run total graupel is handled under the acfrozr variable in default_specs.yml).

Here are a couple of before/after plots for the title adjustment:
image
image

These two plots are also an example of the effect of a long title.  For the radiation flux plots, "Surface" doesn't cause a problem, but "Top of Atmosphere" does, which is why I abbreviated Shortwave and Longwave for these plots, and did that also for the Surface plots to be consistent.  If we can abbreviate "Top of Atmosphere" to "ToA" I can revert the SW and LW abbreviations.

Because the mean RH plot involves many levels, but only plots one wind level, providing it explicitly in the title seems to be the best solution.  Here's an example plot:
image

PR #108, opened 2021-08-18
RRFS smoke update

cshartsough:
This is to update the algorithm (**gribdata**) for Wildfire Potential (firewx) and add it to the **image_list** for RRFS NA 3km runs.

Also made a change to the units in **default_specs**.

passed pylint and pytest.

Sample of latest image.
image

### DIFF
--- a/image_lists/rrfs_subset.yml
+++ b/image_lists/rrfs_subset.yml
@@ -44,6 +44,8 @@ hourly:
       - 2m
     echotop:
       - sfc
+    firewx:
+      - sfc
     flru:
       - sfc
     G113bt:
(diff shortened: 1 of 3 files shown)


image_lists/rrfs_subset.yml
```diff
@@ -92,7 +94,7 @@ hourly:
       - sfc
     ptmp:
       - 2m
-    ptyp
+    ptyp:
```
cshartsough, line 97: I noticed this typo while adding firewx to the file.

Brian-Jamison: I guess that was my fault.  Sorry.  Thanks for fixing it.

Brian-Jamison (approved):
Looks good!

christinaholtNOAA (approved):
Looks good! Thanks!

PR #109, opened 2021-08-31
Resolution fixes for HRRR-CAR and HRRR-HI

Brian-Jamison:
HRRR-Hawaii, and the subdomains for HRRR-Caribbean (also added in this PR), needed settings for higher resolution, since some smaller islands were not being plotted.

### DIFF
--- a/adb_graphics/figures/maps.py
+++ b/adb_graphics/figures/maps.py
@@ -39,10 +39,12 @@ TILE_DEFS = {
     'CHI-DET': [39, 44, -92, -83],
     'DCArea': [36.7, 40, -81, -72],
     'EastCO': [36.5, 41.5, -108, -101.8],
+    'Florida': [19.2305, 29.521, -86.1119, -73.8189],
     'GreatLakes': [37, 50, -96, -70],
     'HI': [16.6, 24.6, -157.6, -157.5],
     'Juneau': [55.741, 59.629, -140.247, -129.274],
     'NYC-BOS': [40, 43, -78.5, -68.5],
+    'PuertoRico': [15.5257, 24.0976, -74.6703, -61.848],
     'SEA-POR': [43, 50, -125, -119],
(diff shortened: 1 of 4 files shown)


create_graphics.py
```diff
@@ -426,6 +426,7 @@ def parallel_maps(cla, fhr, ds, level, model, spec, variable, workdir,
         airport_fn=AIRPORTS,
         ax=ax,
         grid_info=field.grid_info,
+        model_name=cla.model_name,
```
christinaholtNOAA, line 429: The `cla.model_name` is not used in this fashion. It is meant to be whatever string you might like to put in your map title. Apologies that I didn't catch that in PR #103, but when we are requiring a specific string to enable a feature for a given model, we must be in control of that string. A user could run this with "Experiment 2 HRRR-HI" and it wouldn't work appropriately.

The way we've done it for other models is set the `model` key in the relevant image list, which ends up stored in `cla.images[0]`.

adb_graphics/figures/maps.py
```diff
@@ -73,7 +75,7 @@ class Map():
                         TILE_DEFS dictionary
     '''
 
-    def __init__(self, airport_fn, ax, **kwargs):
+    def __init__(self, airport_fn, ax, model_name=None, **kwargs):
```
christinaholtNOAA, line 78: It is likely best to include this as a keyword argument if it's not required. An entry in the documentation block above would also be helpful.

christinaholtNOAA (changes requested):

cshartsough (approved):
good job with the fixes and overall work.

christinaholtNOAA (approved):
Thanks for the super quick response to my earlier comments!

Brian-Jamison:
Thanks!  I forgot to pylint and pytest last time, so hopefully I pass the checks this time.

PR #110, opened 2021-09-01
Fix for RRFS and RTMA_dev1 (RRFS-based) skewT plots (hydrometeors and longitudes)

cshartsough:
This set of changes is to correct problems with  RRFS and  RRFS-based RTMA_dev1 skewT  plotting.  Two issues are corrected
- longitudes were not consistent for -180 to +180 versus 0 to 360.  Checks were added to adjust either way.
- The skewT  code crashed if any of the  hydrometeor fields were not in the grib file,  and cloud water is not currently in the files.  The code was set to skip over missing fields and continue.

passed pylint, pytest needs to be checked (my own local pytest has a big).

sample plots:
image1
image2
999_74001_skewt_f000

### DIFF
--- a/adb_graphics/datahandler/gribdata.py
+++ b/adb_graphics/datahandler/gribdata.py
@@ -259,6 +259,8 @@ class UPPData(specs.VarSpec):
         '''
 
         lats, lons = self.latlons()
+        adjust = 360 if np.any(lons < 0) else 0
+        lons = lons + adjust
         max_x, max_y = np.shape(lats)
 
         # Numpy magic to grab the X, Y grid point nearest the profile site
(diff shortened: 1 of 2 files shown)


Brian-Jamison (approved):
LGTM!

adb_graphics/datahandler/gribdata.py
```diff
@@ -852,10 +854,12 @@ def __init__(self, ds, loc, short_name, **kwargs):
...
         self.site_name = loc[37:].rstrip()
 
-        # Convert the string to a number. Longitude should be negative for all
+        # Convert the string to a number. Longitude should be positive for all
         # these sites.
         self.site_lat = float(lat)
-        self.site_lon = -float(lon)
+        self.site_lon = -float(lon) # lons are -180 but without minus sign in input file
+        if self.site_lon < 0:
+            self.site_lon = self.site_lon + 360.0
```
christinaholtNOAA, line 862: I'm still a little lost with this comment (the original wasn't very clear either).

Let's make sure I'm getting this right. The input files (e.g., conus_raobs.txt) have a positive value for longitude. Here, it was being made negative to be consistent with maybe the specification of the grid in loaded by gribdata.py. That's a bad assumption to make for ALL our models (the rule was broken for RRFS NA), so now, we are ensuring that all data is converted to positive longitude in gribdata.py, and similarly converting profile locations to positive longitudes here, too.

cshartsough: I think my comment in the code got corrupted.

The conus_raobs file uses -180 to 180, but leaves off the minus sign, i.e., the values are  in degrees West.   So, first we need to add the minus sign to convert the longitude to deg East, and then need to adjust to the 0 to 360 system.

christinaholtNOAA (approved):

christinaholtNOAA (approved):
The updated comment is much clearer! Thanks!

PR #112, opened 2021-09-09
Changes needed for RRFS NA 3km

christinaholtNOAA:
Allows the wait time and file age to be specified via command line arguments. Default is set to the times used for HRRR/CONUS domains, while the RRFS NA will need to wait 15 minutes between UPP runs and files should be 5 minutes old (settings for hot fix applied to the real-time run).

Also fixes issues with the naming of variables for RRFS when accumulating over time, especially a fix that allows forecasts to accumulate beyond 6 hours (grib variable LRGHR).

### DIFF
--- a/adb_graphics/datahandler/gribfile.py
+++ b/adb_graphics/datahandler/gribfile.py
@@ -68,8 +68,7 @@ class GribFiles():
 
         self.contents = self._load(filenames)
 
-    @staticmethod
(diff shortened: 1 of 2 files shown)


cshartsough (approved):
nicely done, a lot of hard work and useful upgrades.

Brian-Jamison (approved):
Great job!  Thanks for including the detailed comments.

PR #114, opened 2021-09-15
Explicitly set type on new command line args.

christinaholtNOAA:
It's a good idea to set these explicitly, but also necessary when calling from a bash script. 

This change was already applied as a hot fix in real time for RRFS NA 3km.


cshartsough (approved):
thanks for the fix.

Brian-Jamison (approved):
Thanks, Christina!

PR #115, opened 2021-09-15
Adding a code owners file.

christinaholtNOAA:
This file will automatically add the 3 of us as reviewers on each PR that is opened.


cshartsough (approved):
Thanks for making us all default reviewers... I think.  ;-)

Brian-Jamison (approved):
Thanks!  One less step in a PR!

PR #119, opened 2021-09-27
Memory leak fixes

christinaholtNOAA:
This set of changes fixes a memory leak issue we'd been seeing that does not allow running subdomains on smaller Jet partitions (sjet/vjet and tjet/ujet). Summary of major changes:

- Removed a lot of caching as parallel mode doesn't really use that allocated memory anyway.
- Explicitly cleaned up larger memory objects.
- Clear and close images explicitly once saved.

These changes have been tested for HRRR and RRFS CONUS on a variety of jet partitions with desirable results, most notably the completion of subdomains, even several subdomains.

### DIFF
--- a/adb_graphics/conversions.py
+++ b/adb_graphics/conversions.py
@@ -11,19 +11,19 @@ def k_to_c(field: np.ndarray, **kwargs) -> np.ndarray:
 
     ''' Conversion from Kelvin to Celsius '''
 
-    return np.asarray(field) - 273.15
+    return field - 273.15
 
 def k_to_f(field: np.ndarray, **kwargs) -> np.ndarray:
 
     ''' Conversion from Kelvin to Farenheit '''
 
-    return (np.asarray(field) - 273.15) * 9/5 + 32
+    return (field - 273.15) * 9/5 + 32
 
 def kgm2_to_in(field: np.ndarray, **kwargs) -> np.ndarray:
 
     ''' Conversion from kg per m^2 to inches '''
 
-    return np.asarray(field) * 0.03937
+    return field * 0.03937
 
 def magnitude(a: np.ndarray, b: np.ndarray, **kwargs) -> np.ndarray:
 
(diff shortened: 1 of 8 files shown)


adb_graphics/conversions.py
```diff
@@ -7,91 +7,91 @@
 
 import numpy as np
 
-def k_to_c(field: np.ndarray, **kwargs) -> np.ndarray:
```
christinaholtNOAA, line 10: @cshartsough brought up the issue of some variables being numpy arrays, while others were xarrays. I've removed all instances I could find of converting an array to a numpy array.

adb_graphics/datahandler/gribdata.py
```diff
@@ -473,7 +479,9 @@ def aviation_flight_rules(self, values, **kwargs) -> np.ndarray:
         flru = np.where((ceil > 0.0) & (ceil < 0.5), 3.01, flru)
         flru = np.where((vis < 1.), 3.01, flru)
 
-        return flru
+        vis.close()
+
+        return xr.DataArray(flru)
```
christinaholtNOAA, line 484: The numpy "where" statements above returned a numpy array. Just making sure here that we do get an XArray

adb_graphics/datahandler/gribdata.py
```diff
@@ -530,11 +538,11 @@ def fire_weather_index(self, values, **kwargs) -> np.ndarray:
...
-        temp = self.values(name='temp', level='2m', do_transform=False).values
-        dewpt = self.values(name='dewp', level='2m', do_transform=False).values
-        weasd = self.values(name='weasd', level='sfc', do_transform=False).values
-        gust = self.values(name='gust', level='10m', do_transform=False).values
-        soilm = self.values(name='soilm', level='sfc', do_transform=False).values
+        temp = self.values(name='temp', level='2m', do_transform=False)
+        dewpt = self.values(name='dewp', level='2m', do_transform=False)
+        weasd = self.values(name='weasd', level='sfc', do_transform=False)
+        gust = self.values(name='gust', level='10m', do_transform=False)
+        soilm = self.values(name='soilm', level='sfc', do_transform=False)
```
christinaholtNOAA, line 545: Since all are now Xarrays, we shouldn't need to use the "values" method to get the actually data.

adb_graphics/datahandler/gribdata.py
```diff
@@ -559,9 +567,14 @@ def fire_weather_index(self, values, **kwargs) -> np.ndarray:
...
 
+        temp.close()
+        dewpt.close()
+        weasd.close()
+        gust.close()
+        soilm.close()
+
         return fwi
 
-    @property
```
christinaholtNOAA, line 564: I think that object properties also save the data structure in memory until the object is deleted, so this was another way to allow the freeing of memory.

adb_graphics/datahandler/gribfile.py
```diff
@@ -141,6 +139,7 @@ def _get_grid_suffix(filenames):
         for files in filenames.values():
             if files:
                 gfile = xr.open_dataset(files[0],
+                                        cache=False,
```
christinaholtNOAA, line 142: If we turn off caching here, we may see less memory usage as we reference more and more fields. The docs say this about cached=True, the defualt:

> If True, cache data loaded from the underlying datastore in memory as NumPy arrays when accessed to avoid reading from the underlying data- store multiple times.

adb_graphics/figures/maps.py
```diff
@@ -281,60 +282,11 @@ def draw(self, show=False): # pylint: disable=too-many-locals, too-many-branches
 
         # Contour secondary fields, if requested
         if self.contour_fields:
-            for contour_field in self.contour_fields:
```
christinaholtNOAA, line 284: All this is related to linting and having too many statements in the "draw" method.

adb_graphics/specs.py
```diff
@@ -44,7 +43,6 @@ def centered_diff(self):
         return colors
 
     @property
-    @lru_cache()
```
christinaholtNOAA, line 47: No real need to cache any of this.

tests/test_common.py
```diff
@@ -33,7 +33,6 @@ def test_conversion():
     list, or int '''
 
     a = np.ones([3, 2]) * 300
-    b = list(a)
```
christinaholtNOAA, line 36: Removing support to pass a plain list of values to the conversions methods.

tests/test_grib.py
```diff
@@ -33,7 +35,7 @@ def values(self, level=None, name=None, **kwargs):
         assert isinstance(upp.clevs, np.ndarray)
         assert isinstance(upp.date_to_str(datetime.datetime.now()), str)
         assert isinstance(upp.fhr, str)
-        assert isinstance(upp.field, xarray.DataArray)
+        assert isinstance(upp.field, DATAARRAY)
```
christinaholtNOAA, line 38: We shouldn't be getting back any numpy arrays from fields after this change...

christinaholtNOAA (commented):
Just a few helpful comments as to why I made some of these changes.

cshartsough (approved):
pretty much followed all the changes, and why they can help with memory leaks.  Would be good to review them together at a web graphics meeting, just to learn more about "best practices" as we write new modules and edit the current code down the road. :-)

christinaholtNOAA:
Excellent suggestion, @cshartsough. I will add a card to remind us to do that.

Brian-Jamison (commented):
Great work!  Thanks for all the comments, they are very helpful!
I'm curious if we even need to import numpy (at least, for conversions.py and gribdata.py)?

christinaholtNOAA:
The linter checks for unused imports, so we should be covered there.

Brian-Jamison (approved):
Great work!  Thanks for all the comments, they are very helpful!
I'm curious if we even need to import numpy (at least, for conversions.py and gribdata.py)?

Brian-Jamison:
Whoops, left my previous comment and just 'approved', sorry for the repeat.

Thanks for the clarification on the linter!

PR #120, opened 2021-09-27
Use 2m Temp colors for Soil Temp

christinaholtNOAA:
A request for higher-resolution soil temperature colors came in last week from Tanya. A hot fix was applied for HRRR and RRFS comparisons. She was happy with the result for her comparisons. Putting those changes in for permanence here.

### DIFF
--- a/adb_graphics/default_specs.yml
+++ b/adb_graphics/default_specs.yml
@@ -1086,11 +1086,11 @@ soilm: # Soil Moisture Availability
     unit: "%"
 soilt: # Soil Temperature
   0cm: &soilt_levs
-    clevs: !!python/object/apply:numpy.arange [240, 311, 10]
-    cmap: gist_ncar
-    colors: soilt_colors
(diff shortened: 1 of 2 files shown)


cshartsough (approved):
looks good, a couple small  thoughts:
1. Tanya had asked/suggested to have ticks=8, did that change, or not important?
2. if we're not using soilt_colors, it could be removed from the  specs.py file.
Thanks, good quick work.

Brian-Jamison (approved):
Other than what Craig mentioned, LGTM!  Thanks, nice job!

christinaholtNOAA:
@cshartsough Excellent suggestion for #2. I will do that shortly.

For #1,  I'm not sure the tick marks were important. She returned a little later to say "18z plots (Sept 24) of soil temperature look great, including the tick marks in NA RRFS." I decided just to continue with what we had in place for that time...and those were these changes.

PR #121, opened 2021-09-27
fix bug to allow all skewT info boxes to be plotted.

cshartsough:
This is a one-line fix to plot the hydrometeor legend plot.  The plot call was overwriting the hodograph plot settings.  The corrected call is based on the hydrometeor plotting coordinates so it doesn't overwrite.

passed pylint. pytest failed at my end, I think it may work in GitHub.

sample before and after plots below.

image
image


christinaholtNOAA (approved):
Nice!

PR #123, opened 2021-10-06
correction to firewx to properly use soil moisture

cshartsough:
This change is to use soilw rather than soilm in the Hourly Wildfire Potential (firewx) calculation.  The current code also had an error, rteating the input as a percentage rather than a fraction.

passed pylint.  will check pytest in GitHub.

Before and after for RRFS test (Eric James reviewed):
firewx_full_sfc_f002_before
firewx_full_sfc_f002_after


cshartsough:
> Was there a unit change at all?

Good question.  No, the final value for "mois" is a fraction in both cases.

Issue #111, opened 2021-09-08
Problems with missing forecast hours


Brian-Jamison:
Deployment of HRRR-CAR revealed that having a missing forecast hour causes an index slicing error, and the job fails and restarts until the deadline time set in the rocoto xml.

We need to add an exception to "ignore" this hour for most plots, and either stop any processing of run-total plots at that hour, or add some text to the plots noting any missing hours.

Issue #94, opened 2021-06-03
Create option for continuing real-time graphics after corrupt grib2 file


christinaholtNOAA:
An occasional failure mode exists when a single forecast hour has a corrupt grib2 file. The graphics should detect this sort of failure and continue to create graphics in real-time mode when this happens.

Issue #85, opened 2021-05-05
Run total graphics should not be made when all forecasts are not present


christinaholtNOAA:
Need some checking on dataset to ensure all forecast hours are available before producing a run total graphics.

Issue #84, opened 2021-05-05
Cannot run single forecast hour with `--all_leads` flag


christinaholtNOAA:
The problem was first identified when testing PR #83. Failure was an issue combining the free_fcst files in one big chunk:

```ValueError: 'APCP_P8_L1_GLC0_acc2h' is not present in all datasets.```

This does not seem to be an issue when loading one file at a time.

PR #126, opened 2021-11-01
Combine prs and nat grib files.

christinaholtNOAA:
This PR adds options to concatenate grib files when multiple paths and file templates are provided. When combined, the resulting file will also remove all duplicated entries. The script self-cleans those combined files upon exit. 

This work also addresses what should happen when a forecast hour is not found. If a grib input file is not found within the waiting period, all accumulated variables will be removed from the subsequent forecast hours, and all other images should be generated. 

The feature that allows accumulated variables to be plotted a single forecast lead time (an -all_leads flag with -f 6, for example), should now work as expected. If combining files in this mode, all files must be present. A real-time feature to wait on the file has been disabled since this this is not how real-time cases are run.

Should address Issues #94, #111, #85, and #84.

### DIFF
--- a/create_graphics.py
+++ b/create_graphics.py
@@ -78,7 +78,6 @@ def create_maps(cla, fhr, gribfiles, workdir):
         with Pool(processes=cla.nprocs) as pool:
             pool.starmap(parallel_maps, args)
 
-
 def create_zip(png_files, zipf):
 
     ''' Create a zip file. Use a locking mechanism -- write a lock file to disk. '''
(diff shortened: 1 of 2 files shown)


create_graphics.py
```diff
@@ -108,37 +113,29 @@ def create_zip(png_files, zipf):
...
-        first_fcst = 6 if cla.images[0] == 'global' else 1
-        if fcst_hour <= first_fcst:
-            filenames['01fcst'].append(filename)
-        else:
-            filenames['free_fcst'].append(filename)
+    first_fcst = 6 if cla.images[0] == 'global' else 1
+    if fcst_hour <= first_fcst:
+        filenames['01fcst'].append(filename)
+    else:
+        filenames['free_fcst'].append(filename)
```
christinaholtNOAA, line 130: This code is now only handling one forecast hour at a time. The loop should take place in the main loop for handling missing files appropriately.

create_graphics.py
```diff
@@ -530,6 +535,198 @@ def parallel_skewt(cla, fhr, ds, site, workdir):
...
+                msg = f'graphics: {variable} {level}'
+                raise errors.NoGraphicsDefinitionForVariable(msg)
+            accumulate = spec.get('accumulate', False)
+
+            if accumulate:
+                print(f'Will not plot {variable}:{level}')
+                cla.images[1][variable].remove(level)
+                if not cla.images[1][variable]:
+                    del cla.images[1][variable]
+
```
christinaholtNOAA, line 642: I feel slightly icky about this implementation where I'm changing input arguments explicitly. I may want to clean that up more in a subsequent PR. I think it works okay for now. Happy to hear feedback.

create_graphics.py
```diff
@@ -530,6 +535,198 @@ def parallel_skewt(cla, fhr, ds, site, workdir):
...
+    # Prepare template with all viable forecast hours -- glob accepts *
+    combined_fn = COMBINED_FN.format(fhr=999).replace('999', '*')
+    combined_fp = os.path.join(cla.output_path, combined_fn)
+
+    combined_files = glob.glob(combined_fp)
+
+    print(f'Removing combined files: ')
+    for file_path in combined_files:
+        print(f'  {file_path}')
+        os.remove(file_path)
```
christinaholtNOAA, line 657: Since this needs to be done in a couple of places, using a function.

create_graphics.py
```diff
@@ -530,6 +535,198 @@ def parallel_skewt(cla, fhr, ds, site, workdir):
...
+    for tile in tiles:
+        tile_zip_dir = os.path.join(zip_dir, tile)
+        tile_zip_file = os.path.join(tile_zip_dir, 'files.zip')
+        print(f"checking for {tile_zip_file}")
+        if os.path.isfile(tile_zip_file):
+            os.remove(tile_zip_file)
+            print(f"{tile_zip_file} found and removed")
+        os.makedirs(tile_zip_dir, exist_ok=True)
+        zipfiles[tile] = tile_zip_file
+    return zipfiles
```
christinaholtNOAA, line 685: This was just moved from the driver function after complaints from the linter.

create_graphics.py
```diff
@@ -530,6 +535,198 @@ def parallel_skewt(cla, fhr, ds, site, workdir):
...
+    '''
+
+    for tile, zipf in zipfiles.items():
+        png_files = glob.glob(os.path.join(workdir, f'*_{tile}_*{fhr:02d}.png'))
+        zip_proc = Process(group=None,
+                           target=create_zip,
+                           args=(png_files, zipf),
+                           )
+        zip_proc.start()
+        zip_proc.join()
```
christinaholtNOAA, line 729: Also moved due to linter complaints.

create_graphics.py
```diff
@@ -567,39 +755,65 @@ def graphics_driver(cla):
...
-            print((('-' * 80)+'\n') * 2)
-            print()
-            print(f'Graphics will be created for input file: {grib_path}')
-            print(f'Output graphics directory: {workdir}')
-            print()
-            print((('-' * 80)+'\n') * 2)
+            print(f'{LOG_BREAK}\n',
+                  f'Graphics will be created for input file: {grib_path}\n',
+                  f'Output graphics directory: {workdir} \n'
+                  f'{LOG_BREAK}')
```
christinaholtNOAA, line 811: Using this syntax because linter was complaining about "too many statements"

create_graphics.py
```diff
@@ -32,6 +33,11 @@
 
 AIRPORTS = 'static/Airports_locs.txt'
 
+COMBINED_FN = 'combined_{fhr:03d}.grib2'
+TMP_FN = 'combined_{fhr:03d}.tmp.grib2'
```
christinaholtNOAA, line 37: Using parameters here because multiple methods need to know how these temporary files are named.

christinaholtNOAA (commented):
It may be able to see the diffs in this one with the "split" view in GitHub.

cshartsough (approved):
These changes seem straightforward, I didn't see any problems.  One minor check:
line 126 has `first_fcst = 6 if cla.images[0] == 'global' else 1`, 
line 762 has `first_fcst = 6 if cla.images[0] == 'global' else 0`,

I assume that's correct but just making sure.
   
Great job, as always.

christinaholtNOAA:
Great question! 

The first instance (`first_fcst = 6 if cla.images[0] == 'global' else 1`) is using the following logic, and we want to catch 0, 1, and 6 for the relevant cases:
```
if fcst_hour <= first_fcst:
        filenames['01fcst'].append(filename)
```


And then the second instance (`first_fcst = 6 if cla.images[0] == 'global' else 0`) has this logic which should start the range from either 0 or 6 depending on the model:
```
for fhr in range(first_fcst, int(cla.fcst_hour[0]), fcst_inc):
```


I think that's right...

Brian-Jamison (approved):
Great job!  A fairly heavy lift, thanks for taking it on.  Everything looks good.  One thing to note is the typical schedule of the data.  For a HRRR run, the nat files start coming in about 1:03 hrs after the init time, and come in chunks every 10 minutes.  The prs files start coming in about 53 minutes after the init time, and new forecasts come in about every 2-3 minutes, but some can straggle and come in several minutes later.  So some adjustment may be needed for the wait times when combining prs and nat files.

christinaholtNOAA:
@Brian-Jamison That's a great point! I think we will really need to consider how we want to wait and submit these jobs. The code will wait for a while to make sure the real-time files are available, but maybe we want to delay the start of the runs until nat files start showing up.

PR #127, opened 2021-11-03
correct field range and color table for uswrf

cshartsough:
The original default settings for upward shortwave radiation flux (uswrf) were copied from the longwave values, which is not correct.  I changed the range and color table to approximate the old NCL settings.

passed pylint and pytest on local machine.

sample before and after below.
image
image


christinaholtNOAA (approved):
LGTM!

Brian-Jamison (approved):
Looks good!  Thanks, Craig!

PR #128, opened 2021-11-03
Attempting to speed up conda install.

christinaholtNOAA:
First try...an attempt to only use conda-forge after reading through suggestions on this issue: https://github.com/conda/conda/issues/7690

Second try...set strict priority channel setting in condarc.


christinaholtNOAA:
Putting this on the back burner for now.

PR #129, opened 2021-11-03
update RRFS soil levels in image_list

cshartsough:
This resets the RRFS soil levels to the same 9 values as the HRRR.  It is in preparation for the upcoming  deployment which will change the levels from 4 to 9.

Testing using Eric's test file data.

Passed pylint and pytest.  No graphics comparison needed.


Brian-Jamison (approved):
Thanks for handling this, Craig!

PR #130, opened 2021-11-19
Apply hot fix for SkewTs

christinaholtNOAA:
This fix was needed and was already deployed in real time as a hot fix. A new set of logic was only applicable for accumulating fields for maps.


cshartsough (approved):
Looks like this change is just to move the block into a conditional, no other changes?  Seems like it should be fine, based on my new training in Best Practices.

Brian-Jamison (approved):
LGTM!

PR #131, opened 2021-11-19
change color table to improve soilw comparisons

cshartsough:
Tanya requested a change to the soilw color table to allow comparisons to other models.

Before and after sample plots below.

Passed pylint and pytest.

image
image

### DIFF
--- a/adb_graphics/default_specs.yml
+++ b/adb_graphics/default_specs.yml
@@ -1124,8 +1124,8 @@ soilt: # Soil Temperature
     title: Soil Temperature at 3m
 soilw: # Soil Moisture
   0cm: &soilw_levs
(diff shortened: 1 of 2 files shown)


christinaholtNOAA (approved):
This looks good to me from the code standpoint. We're probably awfully close to that yellow/green transition problem.

Brian-Jamison (approved):
Looks good, thanks!

cshartsough:
> 

true, we need to watch our colors; in this case, as with visibility, the request was to specifically match the plots from an outside group, to make comparisons easier.  We might need to think about how to handle that type of request down the road.

Brian-Jamison (commented):
Those dreaded spaces! ;-)

Brian-Jamison (commented):
LGTM!

PR #132, opened 2021-11-19
change title for 10m Gust, include "Potential"

cshartsough:
Stan requested a change to the 10m wind gust plot title, to make sure it included "Potential".

sample before and after plots below.

passed pylint and pytest

image

image


Brian-Jamison (approved):
Thanks, Craig!

PR #133, opened 2021-11-22
sync snow color tables, adjust yellow/orange

cshartsough:
Ed Szoke requested two small changes for snow plots:
(1) sync so that all snow depth use the same color table. changed 1h "snod" vs full run "acsnod" accumulated, based on his preference.
(2) change the yellow color slightly to make levels clearer for colorblindness.

sample before and after shown below, for plots (old, then new) and colorblind perspective (new on top, old below).

passed pylint and pytest.

image

image

Screen Shot 2021-11-21 at 9 28 58 PM
)


Brian-Jamison (approved):
LGTM!

PR #134, opened 2021-11-23
Use random strings in tmp file names.

christinaholtNOAA:
When running in real time, we run several jobs simultaneously, split over subdomains in unpredictable ways.

Add a random string to the temporary file names to create thread safety.


cshartsough (approved):
looks okay.  my code is more random than this.  ;-)

Brian-Jamison (approved):
LGTM!

PR #135, opened 2021-11-23
adjust blue end of soilw colors for more detail

cshartsough:
Tanya asked for a small change in the soilw color table, to make some of the greenish-blues more blue.  

Sample change below.  Will share it with Tanya in Slack while the GitHub test is running, and redo if she wants more changes.

passed pylint and pytest.

Screen Shot 2021-11-23 at 12 26 42 PM


Brian-Jamison (approved):
I'm good with it.  I hope Tanya is, also.

christinaholtNOAA (approved):
Looks fine. Sigh.

PR #137, opened 2021-12-08
Global plots part2

Brian-Jamison:
This is part 2 of the global plots, which cover the interpolated subregions for Alaska, CONUS, and the Northern and Southern Hemisphere.  The Southern Hemisphere cannot currently be plotted correctly with the map, because the metadata in the grib2 file is incorrect.

Also included here is the ability to turn off airport plotting for certain plots.  In this case, the airports for the smoke-related plots are not plotted (by request).

Examples of the new imagery follows:

Alaska full domain:
image

Alaska Zoom:
image

Alaska Zoom 2:
image

CONUS full:
image

NW:
image
 
 NC:
image

NE:
image

SW:
image

SC:
image

SE:
image

NHemi:
image

### DIFF
--- a/adb_graphics/datahandler/gribdata.py
+++ b/adb_graphics/datahandler/gribdata.py
@@ -616,8 +616,6 @@ class fieldData(UPPData):
         elif self.grid_suffix == 'GLL0':
             attrs = []
             grid_info['projection'] = 'cyl'
-            grid_info['lon_0'] = 0
-            grid_info['lat_0'] = 90
         else:
             attrs = []
             grid_info['projection'] = 'rotpole'
--- a/adb_graphics/datahandler/gribdata.py
+++ b/adb_graphics/datahandler/gribdata.py
@@ -39,7 +39,6 @@ class UPPData(specs.VarSpec):
         # Parse kwargs first
         config = kwargs.get('config', 'adb_graphics/default_specs.yml')
         self.model = kwargs.get('model')
-        self.tile = kwargs.get('tile')
         self.filetype = kwargs.get('filetype', 'prs')
 
 
@@ -278,7 +277,6 @@ class UPPData(specs.VarSpec):
 
         if x <= 0 or y <= 0 or x >= max_x or y >= max_y:
             print(f'site location is outside your domain! {site_lat} {site_lon}')
-            # return(-1.E10, -1.E10)
             return(-1, -1)
 
         return (x, y)
@@ -527,10 +525,7 @@ class fieldData(UPPData):
         if self.model == 'global':
             ret = [lat[-1], lat[0], lon[0], lon[-1]]
         else:
-            ret = [lat[0, 0], lat[-1, -1], lon[0, 0], lon[-1, -1]]  # good SHemi data but wrong map
-            #ret = [20.826, 20.826, 30.0, 120.0]
-            #ret = [20.826, 20.826, 120, -60]  # good SHemi map but no data
-        print(f'ret = {ret}')
+            ret = [lat[0, 0], lat[-1, -1], lon[0, 0], lon[-1, -1]]
 
         return ret
 
@@ -616,22 +611,13 @@ class fieldData(UPPData):
             grid_info['lat_0'] = 39.0
         elif self.grid_suffix == 'GST0':
             attrs = ['Lov']
-            grid_info['projection'] = 'stere'  # don't use npstere or spstere
-            #grid_info['lat_0'] = 90  # works for NHemi
-            #grid_info['lat_0'] = -90  # works for SHemi
-            grid_info['lat_ts'] = 90
+            grid_info['projection'] = 'stere'
             grid_info['lat_0'] = 90
         elif self.grid_suffix == 'GLL0':
             attrs = []
-            #grid_info['projection'] = 'cyl'
-            grid_info['projection'] = 'ortho'
-            #grid_info['resolution'] = None
+            grid_info['projection'] = 'cyl'
             grid_info['lon_0'] = 0
             grid_info['lat_0'] = 90
-            grid_info['llcrnrx'] = 0
-            grid_info['llcrnry'] = 0
-            grid_info['urcrnrx'] = 1439
-            grid_info['urcrnry'] = 720
         else:
             attrs = []
             grid_info['projection'] = 'rotpole'
--- a/adb_graphics/datahandler/gribdata.py
+++ b/adb_graphics/datahandler/gribdata.py
@@ -547,7 +547,7 @@ class fieldData(UPPData):
         dewpt = self.values(name='dewp', level='2m', do_transform=False)
         weasd = self.values(name='weasd', level='sfc', do_transform=False)
         gust = self.values(name='gust', level='10m', do_transform=False)
-        soilm = self.values(name='soilm', level='sfc', do_transform=False)
+        soilw = self.values(name='soilw', level='0cm', do_transform=False)
 
         # A few derived fields
         dewpt_depression = temp - dewpt
@@ -556,7 +556,7 @@ class fieldData(UPPData):
         snowc = (25.0 - weasd) / 25.0
         snowc = np.where(snowc > 0.0, snowc, 0.0)
 
-        mois = 0.01 * (100.0 - soilm)
+        mois = 1.0 - soilw
 
         # Set urban (13), snow/ice (15), barren (16), and water (17) to 0.
         for vegtype in [13, 15, 16, 17]:
@@ -576,7 +576,7 @@ class fieldData(UPPData):
         dewpt.close()
         weasd.close()
         gust.close()
-        soilm.close()
+        soilw.close()
 
         return fwi
 
--- a/adb_graphics/datahandler/gribdata.py
+++ b/adb_graphics/datahandler/gribdata.py
@@ -267,8 +267,6 @@ class UPPData(specs.VarSpec):
         lats, lons = self.latlons()
         adjust = 360 if np.any(lons < 0) else 0
         lons = lons + adjust
-        #if self.model in ['global']:
-        #    lats, lons = np.meshgrid(lats, lons, sparse=False, indexing='ij')
         max_x, max_y = np.shape(lats)
 
         # Numpy magic to grab the X, Y grid point nearest the profile site
@@ -524,11 +522,10 @@ class fieldData(UPPData):
         '''
 
         lat, lon = self.latlons()
-        ret = [lat[0, 0], lat[-1, -1], lon[0, 0], lon[-1, -1]]
-        #if self.model == 'global':
-        #    ret = [lat[-1], lat[0], lon[0], lon[-1]]
-        #else:
-        #    ret = [lat[0, 0], lat[-1, -1], lon[0, 0], lon[-1, -1]]
+        if self.model == 'global':
+            ret = [lat[-1], lat[0], lon[0], lon[-1]]
+        else:
+            ret = [lat[0, 0], lat[-1, -1], lon[0, 0], lon[-1, -1]]
 
         return ret
 
--- a/adb_graphics/datahandler/gribdata.py
+++ b/adb_graphics/datahandler/gribdata.py
@@ -512,7 +512,10 @@ class fieldData(UPPData):
         '''
 
         lat, lon = self.latlons()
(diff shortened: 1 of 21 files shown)


adb_graphics/datahandler/gribfile.py
```diff
@@ -94,7 +94,7 @@ def free_fcst_names(self, ds, fcst_type):
                         ])
                 needs_renaming = var.split('_')[0] not in odd_variables
                 if suffix in special_suffixes and needs_renaming:
-                    new_suffix = f'{suffix}1h' if self.model not in ['global'] else f'{suffix}6h'
+                    new_suffix = f'{suffix}1h' if 'global' not in self.model else f'{suffix}6h'
                     ret[var] = var.replace(suffix, new_suffix)
             else:
```
Brian-Jamison, line 99: These were changed to trigger for globalAK, globalCONUS, and globalNHemi.

adb_graphics/default_specs.yml
```diff
@@ -1323,6 +1331,7 @@ trc1:
     clevs: [1, 4, 7, 11, 15, 20, 25, 30, 40, 50, 75, 150, 250, 500]
     colors: smoke_colors
     ncl_name: COLMD_P0_L200_{grid}
+    plot_airports: False
```
Brian-Jamison, line 1334: The new parameter to turn off airports if desired.

adb_graphics/figures/maps.py
```diff
@@ -86,15 +88,16 @@ class Map():
                         TILE_DEFS dictionary
     '''
 
-    def __init__(self, airport_fn, ax, **kwargs):
+    def __init__(self, airport_fn, ax, field, **kwargs):
```
Brian-Jamison, line 91: Need to include field here so we can turn on/off airports.

adb_graphics/figures/maps.py
```diff
@@ -520,30 +527,21 @@ def _wind_barbs(self, level):
...
-            stride = 20
-            length = 4
+                stride = int(round(u.shape[1] / 35))
+            length = 5
         else:
             stride = TILE_DEFS[tile]["stride"]
             length = TILE_DEFS[tile]["length"]
+            if self.map.model == 'globalCONUS':
+                stride = int(round(stride / 2.5))
+                length = 5
```
Brian-Jamison, line 544: A new feature, a decent stride can be found if you divide the number of grid points on the shorter side by 35.  Subdomains are defined by lat,lon so the stride is set in the TILE_DEFS.  For the CONUS subdomains, further dividing by 2.5 works well.

image_lists/global.yml
```diff
@@ -3,12 +3,6 @@ hourly:
   variables:
     1ref:
       - 1000m
-    acpcp:
-      - sfc
-    acsnod:
-      - sfc
-    acsnw:
-      - sfc
```
Brian-Jamison, line 11: Because of the presence of another total acc variable, the renaming has trouble, so these will be left out for now until a fix is put in.

image_lists/global.yml
```diff
@@ -49,10 +43,6 @@ hourly:
       - 10cm
       - 1m
     soilw: *soilt_levs
-    solar:
-      - sfc
-    ssrun:
-      - sfc
```
Brian-Jamison, line 55: These products aren't available for global.

Brian-Jamison:
I put in some comments about the code, and added sample images.

image_lists/globalAK.yml
```diff
@@ -0,0 +1,66 @@
+hourly:
```
christinaholtNOAA, line 1: I see some diffs between image_lists for the 4 global ones. Should they be different? If not, then we may want to consider a different solution to how to handle the different tiles.

christinaholtNOAA (changes requested):
Brian, thanks for the follow up to address some of the global subdomain issues! 

I have some comments/questions/concerns below.

adb_graphics/default_specs.yml
Brian-Jamison: Yes, it is now.  Thanks for reminding me.

adb_graphics/figures/maps.py
Brian-Jamison: Good point in not passing the field.  This has been changed to pass a boolean flag for plot_airports.

adb_graphics/figures/maps.py
Brian-Jamison: Done.

adb_graphics/datahandler/gribdata.py
```diff
@@ -613,10 +613,11 @@ def grid_info(self):
             attrs = ['Lov']
             grid_info['projection'] = 'stere'
             grid_info['lat_0'] = 90
-            grid_info['lat_ts'] = 90
```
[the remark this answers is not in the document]
Brian-Jamison: The documentation on lat_ts states: "latitude of true scale. Optional for stereographic, cylindrical equal area and mercator projections. default is lat_0 for stereographic projection. default is 0 for mercator and cylindrical equal area projections."  So it defaults to lat_0, which is 90, and thus not needed.  It may be that it will need to be set for the Southern Hemisphere plot, when that is available.

adb_graphics/figures/maps.py
```diff
@@ -32,7 +32,8 @@
     'SE': {'corners': [22, 37, -93.5, -72], 'stride': 10, 'length': 4},
     'SW': {'corners': [24.5, 45, -122, -103], 'stride': 10, 'length': 4},
     'Africa': {'corners': [-40, 40, -40, 60], 'stride': 7, 'length': 5},
-    'AKZoom': {'corners': [52, 73, -162, -132], 'stride': 4, 'length': 4},
+    'AKZoom': {'corners': [52, 73, -162, -132], 'stride': 4, 'length': 5},
```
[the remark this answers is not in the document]
Brian-Jamison: Yes, it's been tested with RAP-AK, which is the only other model.

adb_graphics/figures/maps.py
```diff
@@ -144,14 +147,17 @@ def draw_airports(self):
...
-                    ax=self.ax,
-                    color='w',
-                    fillstyle='full',
-                    markeredgecolor='k',
-                    markeredgewidth=0.5,
-                    markersize=4,
-                    )
+
+        add_airports = self.field.vspec.get('plot_airports', True)
+        if add_airports:
```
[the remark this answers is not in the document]
Brian-Jamison: Thanks, corrected.

adb_graphics/figures/maps.py
```diff
@@ -310,8 +316,9 @@ def draw(self, show=False): # pylint: disable=too-many-locals, too-many-branches
             self._wind_barbs(add_wind)
 
         # Add field values at airports
+        add_airports = self.field.vspec.get('plot_airports', True)
         annotate = self.field.vspec.get('annotate', False)
-        if annotate and self.map.model not in ['global']: # airports are too dense in global
+        if add_airports and annotate and 'global' not in self.map.model: # too dense in global
```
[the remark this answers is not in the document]
Brian-Jamison: Good point.  I had that in case we wanted to plot airports for Alaskan (or other) subdomains for global.  If that's wanted at some point, the logic here will need to be changed.

adb_graphics/figures/maps.py
```diff
@@ -375,7 +382,7 @@ def _draw_field(self, ax, field, func, **kwargs):
         '''
 
         x, y = self._xy_mesh(field)
-        vals = field.values()[::]
+        vals = field.values()
```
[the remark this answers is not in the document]
Brian-Jamison: It appears to be unnecessary.  I think I removed it at one time to try and locate the source of an error, and though it wasn't the source, I left the change because it didn't seem to cause an error or have an effect.  I've tested it this way with all the models, and the result is the same with or without it.

image_lists/globalAK.yml
Brian-Jamison: They have been changed to match.  The only difference is the model names, which may need to be handled separately due to different projections.

create_graphics.py
```diff
@@ -445,12 +445,15 @@ def parallel_maps(cla, fhr, ds, level, model, spec, variable, workdir,
 
     fig, ax = plt.subplots(1, 1, figsize=(inches, inches))
 
+    plot_airports = spec.get('plot_airports', True)
```
christinaholtNOAA, line 448: This is a pretty short call to spec. You could put it directly in the Map object creation on line 456 to reduce the number of variables created in this method.

image_lists/globalAK.yml
christinaholtNOAA: Since it's already happened at least once, this sort of large duplication will likely create big problems for us. I'd prefer to see this handled differently. Maybe I can try to get this knocked out in this "down week" so that we can also handle the RRFS domains similarly.

christinaholtNOAA (changes requested):

create_graphics.py
```diff
@@ -451,6 +451,7 @@ def parallel_maps(cla, fhr, ds, level, model, spec, variable, workdir,
         ax=ax,
         grid_info=field.grid_info(),
         model=model,
+        plot_airports=spec.get('plot_airports', True),
```
Brian-Jamison, line 454: The spec get has been moved to be defined in the call.

adb_graphics/figures/maps.py
```diff
@@ -82,6 +84,8 @@ class Map():
                              ll_lat, ur_lat, ll_lon, ur_lon
           model         model designation used to trigger higher resolution maps if needed
                         also used to turn off plotting of airports on global maps
+          plot_airports bool to allow airport plotting to be turned off for
+                        certain plots, default is True
```
Brian-Jamison, line 88: Documentation added.

adb_graphics/figures/maps.py
```diff
@@ -91,10 +95,11 @@ def __init__(self, airport_fn, ax, **kwargs):
         self.ax = ax
         self.grid_info = kwargs.get('grid_info', {})
         self.model = kwargs.get('model')
+        self.plot_airports = kwargs.get('plot_airports', True)
```
Brian-Jamison, line 98: plot_airports has been removed from the __init__ args, and retrieved using kwargs.get with a default of True.

image_lists/globalAK.yml
Brian-Jamison: I agree, there's a lot of duplication.  I would prefer to handle this in a separate PR.

christinaholtNOAA (approved):
@Brian-Jamison thanks for those changes. It all looks good. Let's follow up with a discussion on resolving conflicts and removing the duplicate image lists.

adb_graphics/datahandler/gribdata.py
```diff
@@ -613,10 +613,11 @@ def grid_info(self):
...
         elif self.grid_suffix == 'GLL0':
             attrs = []
             grid_info['projection'] = 'cyl'
+            grid_info['lon_0'] = 0
+            grid_info['lat_0'] = 90
```
christinaholtNOAA, line 620: Are these set in the file's metadata, or no?

adb_graphics/default_specs.yml
```diff
@@ -1345,6 +1352,7 @@ trc1:
     clevs: [1, 2, 4, 6, 8, 12, 16, 20, 25, 30, 40, 60, 100, 200]
     colors: smoke_colors
     ncl_name: MASSDEN_P0_L103_{grid}
+    plot_airports: False
```
christinaholtNOAA, line 1355: Do we want to turn these off for all projections?

christinaholtNOAA (approved):
LGTM, but see my comments/questions below.

cshartsough (approved):
Looks of hard work here.  Thanks Christina for your review; nothing to add.

Congrats.

adb_graphics/datahandler/gribdata.py
Brian-Jamison: I tested with these two lines commented out, relying on the default values.  Everything ran perfectly, and the resulting plots looked the same.  I removed the lines.

adb_graphics/default_specs.yml
Brian-Jamison: Ravan has asserted that he doesn't want any airports plotted on any of the smoke-related maps, so I think this is ok.

PR #141, opened 2021-12-20
adjust hydro table location in skewTs

cshartsough:
putting through a PR for a hot fix, to re-position the table for vertically integrated hydrometeors in the skewT plots.  This was noticed while making ICICLE plots with RRFS but can be used for all skewTs.


Brian-Jamison (approved):
Thanks for doing this!

PR #142, opened 2021-12-20
 adjust wind colors to have less greens

cshartsough:
a small change to the wind_colors and wind_colors_high color tables.  Stan had request that the mid range not have so many greens.  I adjust the lower end greens to be more blue.  Putting the PR through but may still tweak it if Stan has more comments.

sample plots below.

Wind colors test


christinaholtNOAA (approved):
This is a nicer gradient. The code looks fine, too.

PR #143, opened 2021-12-27
add transform for sfc, 1000ft, and 6000ft smoke

Brian-Jamison:
Due to a units change by NCEP, the values in the MASSDEN smoke fields needs to be multiplied by 1e+9 to convert to micrograms per cubic meter.  The new transform is added here.

### DIFF
--- a/adb_graphics/conversions.py
+++ b/adb_graphics/conversions.py
@@ -78,6 +78,12 @@ def to_micro(field, **kwargs):
 
     return field * 1E6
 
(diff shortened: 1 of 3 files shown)


cshartsough (approved):
Looks good.  We'll probably need to review our test_common soon to make sure we have the proper tests for conversions and other pieces we've added.

Brian-Jamison:
Good point!  I've added the conversion test in test_common.py.

christinaholtNOAA (approved):
LGTM!

PR #145, opened 2022-01-05
Support for named "full" domains.

christinaholtNOAA:
Add support to recognize some grids that are wgrib2 cut outs. These named grids should plot the entirety of the data in the file and use the file metadata to decide the basemap, but still need to be differentiated as "sub domains" or "tiles" of a larger grid for the same modeling system. 

Here I am adding "hrrr" and "hrrrak" for consistency with real-time RRFS NA 3km runs.

### DIFF
--- a/adb_graphics/figures/maps.py
+++ b/adb_graphics/figures/maps.py
@@ -20,6 +20,15 @@ import numpy as np
 
 import adb_graphics.utils as utils
 
+# FULL_TILES is a list strings that includes the labels GSL attaches to some of
+# the wgrib2 cutouts used for larger domains like RAP and RRFS NA.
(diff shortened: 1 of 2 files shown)


cshartsough (approved):
Looks nice.  clean changes.  I had to relook at the line about full_tile = tile in FULL_TILES, at first I though it was a loop but a saw it was just seeing if the selected tile was a full one.

Should never review after 4:30 pm.  :-)

Brian-Jamison (approved):
There will be a few conflicts with the global_plots_part2 PR, but I think they won't be too difficult to resolve.  I think this is a good way to handle the differing names of the "full" domains.

PR #149, opened 2022-02-17
Removing airport plotting from Fire Radiative Power plot.

Brian-Jamison:
Forgot to include this in the global plots part2 PR.  Airports need to be removed from the Fire Radiative Power plot.


christinaholtNOAA (approved):

cshartsough (approved):
thanks for taking care of this

PR #151, opened 2022-03-07
Correct Title for Helicity 0-3km Plots

cshartsough:
Ed Szoke noticed while preparing slides that the 0-3km helicity fields are mislabelled as "0-6km".  Fixed in two places in default_specs.

Passed pylint.   No plots provided.


Brian-Jamison (approved):
Thanks, Craig!

christinaholtNOAA (approved):
LGTM!

PR #152, opened 2022-03-21
Set modules to exact versions. Default not supported.

christinaholtNOAA:
This change has been made on disk in many of the HRRR and RAP real time graphics. It will also need to be deployed for RRFS. 

Jet and Hera are going to discontinue support for default modules, and the exact version must be specified to avoid script failure. 

These are the versions of the modules that are listed in the log files.

### DIFF
--- a/pre.sh
+++ b/pre.sh
@@ -3,9 +3,9 @@
 module purge
 
 module use -a /contrib/miniconda3/modulefiles
-module load miniconda3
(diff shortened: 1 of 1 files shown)


cshartsough (approved):
seems pretty straightforward.  fingers crossed. :-)

Brian-Jamison (commented):
LGTM!

PR #155, opened 2022-03-24
add png_suffix for ctop/ceil to fix 48h Web bug

cshartsough:
For unknown reason, the lack of a `png_suffix` for some image names was causing those images to be lost for real time HRRR 48h forecast time. The plots were available and would populate the Web until the 48h file was added before all forecast hours would be lost.

The fields had a unique level, `ua`, which was not added to images for similarity to legacy plot names.  Removing the conditional for `ua` levels adds that suffix to the plots and clears up the issue.

Passed pylint.  No images required.

### DIFF
--- a/create_graphics.py
+++ b/create_graphics.py
@@ -468,7 +468,7 @@ def parallel_maps(cla, fhr, ds, level, model, spec, variable, workdir,
     dm.draw(show=True)
 
     # Build the output path
-    png_suffix = level if level != 'ua' else ''
+    png_suffix = level
     png_file = f'{variable}_{tile}_{png_suffix}_f{fhr:03d}.png'
     png_file = png_file.replace("__", "_")
     png_path = os.path.join(workdir, png_file)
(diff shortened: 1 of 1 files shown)


Brian-Jamison (approved):
LGTM!

create_graphics.py
```diff
@@ -468,7 +468,7 @@ def parallel_maps(cla, fhr, ds, level, model, spec, variable, workdir,
     dm.draw(show=True)
 
     # Build the output path
-    png_suffix = level if level != 'ua' else ''
+    png_suffix = level
     png_file = f'{variable}_{tile}_{png_suffix}_f{fhr:03d}.png'
```
christinaholtNOAA, line 472: Since we no longer need the logic, `png_suffix` is no longer needed, really. Would you mind putting `level` directly into the template for the `png_file` definition. Unless, of course, I missed where `png_suffix` is used elsewhere.

christinaholtNOAA (changes requested):

christinaholtNOAA (approved):
THANK YOU!

PR #156, opened 2022-03-29
Control output image resolution via command line argument

Jeff-Duda:
Modified create_graphics.py to add an optional command line argument 'img_res' to control the resolution of output images [i.e., the dpi entry in pyplot.savefig()].

### DIFF
--- a/create_graphics.py
+++ b/create_graphics.py
@@ -233,9 +233,9 @@ def parse_args():
     parser.add_argument(
         '-r',
         dest='img_res',
-        default=75,
+        default=72,
         required=False,
-        help='Resolution of output images in DPI. Recommended to stay below 1000. Default = 75',
+        help='Resolution of output images in DPI. Recommended to stay below 1000. Default = 72',
         type=int,
         )
     parser.add_argument(
--- a/create_graphics.py
+++ b/create_graphics.py
@@ -391,7 +391,6 @@ def parallel_maps(cla, fhr, ds, level, model, spec, variable, workdir,
       model      model name: rap, hrrr, hrrre, rrfs, rtma
       spec       the dictionary of specifications for the given variable
                  and level
-      imsiz      resolution of output image in DPI units
       variable   the name of the variable section in the specs file
       workdir    output directory
     '''
--- a/create_graphics.py
+++ b/create_graphics.py
@@ -231,6 +231,14 @@ def parse_args():
 
     # Short args
     parser.add_argument(
+        '-r',
+        dest='img_res',
+        default=75,
+        required=False,
+        help='Resolution of output images in DPI. Recommended to stay below 1000. Default = 75',
+        type=int,
+        )
+    parser.add_argument(
         '-a',
         dest='data_age',
         default=3,
@@ -383,6 +391,7 @@ def parallel_maps(cla, fhr, ds, level, model, spec, variable, workdir,
       model      model name: rap, hrrr, hrrre, rrfs, rtma
       spec       the dictionary of specifications for the given variable
                  and level
+      imsiz      resolution of output image in DPI units
       variable   the name of the variable section in the specs file
       workdir    output directory
     '''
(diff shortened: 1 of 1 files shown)


create_graphics.py
```diff
@@ -230,6 +230,14 @@ def parse_args():
         )
 
     # Short args
+    parser.add_argument(
+        '-r',
+        dest='img_res',
+        default=75,
```
christinaholtNOAA, line 236: Can we keep the default at 72? I'm not sure how sensitive the front-end software is to this number.

create_graphics.py
```diff
@@ -531,7 +539,7 @@ def parallel_skewt(cla, fhr, ds, site, workdir):
     plt.savefig(
         png_path,
         bbox_inches='tight',
-        dpi='figure',
+        dpi=cla.img_res,
```
christinaholtNOAA, line 542: I think this would be fine for a line graph (default = 72), but would you mind providing examples in the PR of the before and after, along with the change in size of the resulting PNG files? I have no clue what the 'figure' dpi would be for these images.

christinaholtNOAA (changes requested):

create_graphics.py
Jeff-Duda: Oops! I had actually intended to keep the default at 72, but got the wrong number in my head when changing the code.

create_graphics.py
Jeff-Duda: That's a good question. Has anyone run the skew-t side of this yet? I have a retro sitting around that I will test this on, but as of right now, I do not know. I have never used 'figure' before as a value for dpi in savefig.

christinaholtNOAA: I'm not sure what you mean. We run this in real-time runs constantly. I just don't know what the dpi numerical value would be for a SkewT figure. I'd like to make sure that 72 as a default wouldn't degrade what is currently produce. My feeling is that it wouldn't, but we do need to check.

Jeff-Duda: Okay, so I have looked at this, and the results are pleasing.

Here is a sample skew-t plot (Denver site) using the existing code (DPI = 72); image size = 211 KB:
DNR_72469_skewt_f001

And here is the same image using the updated code with DPI = 150 (image size = 555 KB):
DNR_skewt_r150

"DPI" = "dots per inch" and is a generic characteristic of making images. There is nothing special about the difference between dpi for skewts and dpi for horizontal cross sections; the same value should work just fine for both types of images.

Regarding 'figure' as the value of dpi in savefig, the documentation for matplotlib says:

>     The resolution in dots per inch. If 'figure', use the figure's dpi value.

It turns out that 'dpi' is also a kwarg for pyplot.figure when declaring a new figure. I was not aware of this previously. However, I don't see dpi explicitly set in skewt.py at line 401:

`        fig = plt.figure(figsize=(12, 12))`

So I presume some default value is being used there. I just figured I would change the code for consistency's sake. I can confirm that if 
`dpi='figure'`
is used, then -r is ignored. So I think changing this in both places makes the code more consistent.

christinaholtNOAA: I think this looks totally acceptable, and am happy with the outcome. The text is slightly blurry with the dpi=72 option, but if we get complaints, we can probably just add the appropriate CLA to the workflows that triggered the complaints. I don't think this will break anything.

christinaholtNOAA (approved):

Brian-Jamison (commented):
This is a great addition, Jeff!  Do you have any information on how much extra time it takes to create the higher resolution image?  It would be good to know if we get a request to generate higher res images in real-time.

Jeff-Duda:
> This is a great addition, Jeff! Do you have any information on how much extra time it takes to create the higher resolution image? It would be good to know if we get a request to generate higher res images in real-time.

I don't, Brian, but I could pretty easily run some timing tests to ascertain this information.

Brian-Jamison:
If it's not too much trouble, please add that information in this conversation before merging so we have it as a reference.  Thanks!

Brian-Jamison (approved):

Brian-Jamison:
Thanks for the size and timing info!  That's helpful to know.

Great job!

PR #157, opened 2022-04-20
Add ensemble panel plots to support RRFSE.

christinaholtNOAA:
This reorganizes some of the code to slim-down create_graphics.py to organize data and setup parallelism for the most part. Utilities and graphics-building have been moved out to the adb_graphics package. 

The code now supports plotting 12-panel plots to support 10-member forecast ensembles for RRFSE. It will also allow composite reflectivity observations to be plotted on the bottom left panel for those figures. 

@cshartsough contributed the design aspect to this work, gathering feedback from the interested AVID ensemble folks and providing the prototypes for how the figures would be built. 

Because this is a big structural change, I'll add some comments in-line after I open the PR. 

The code has currently been tested for the ensemble case for the 7 major domains, and the same list of fields plotted as a 9-member ensemble in 2021. I will include examples below.

The tests to ensure single figures are still working are pending.


adb_graphics/default_specs.yml
```diff
@@ -420,10 +423,15 @@ cref: # Composite reflectivity
     <<: *refl
     ncl_name: REFC_P0_L10_{grid}
     title: Ensemble Composite Reflectivity
+  obs:
+    <<: *refl
+    ncl_name: MergedReflectivityQCComposite_P0_L102_GLL0
+    title: Observed Composite Reflectivity
```
christinaholtNOAA, line 429: I thought it might work okay to add the observations as an additional level to plot here. The script handles ensuring we are pulling it from the appropriate user-provided grib file.

adb_graphics/figure_builders.py
```diff
@@ -0,0 +1,280 @@
+# pylint: disable=invalid-name
```
christinaholtNOAA, line 1: This new file is a collection of old functions that build the skewt and map plots, plus new function that helps set the figure parameters and add the observation panel for composite reflectivity.

adb_graphics/figure_builders.py
```diff
@@ -0,0 +1,280 @@
...
+    fig, axes = set_figure(cla.model_name, cla.graphic_type, tile)
+
+    # set last_panel to send into DataMap for colorbar control
+    last_panel = False
+
+    # Declare the type of object depending on graphic type
+    map_class = maps.MultiPanelDataMap if cla.graphic_type == \
+        'enspanel' else maps.DataMap
+
+    for index, current_ax in enumerate(axes):
```
christinaholtNOAA, line 88: This loop allows us to plot 1 or many panels consistent with the figure that is setup above.

adb_graphics/figure_builders.py
```diff
@@ -0,0 +1,280 @@
...
+        dm = map_class(
+            map_fields=map_fields,
+            map_=m,
+            member=mem,
+            model_name=cla.model_name,
+            last_panel=last_panel
+            )
+
+        # Draw the map
+        if cla.graphic_type == 'enspanel':
```
christinaholtNOAA, line 149: It was a design choice to add the title to the blank top left panel (panel 0) and observations for composite reflectivity to the bottom left (panel 8).

adb_graphics/figure_builders.py
```diff
@@ -0,0 +1,280 @@
...
+    plt.cla()
+    # Clear the current figure.
+    plt.clf()
+    # Closes all the figure windows.
+    plt.close('all')
+    del field
+    del m
+    gc.collect()
+
+def parallel_skewt(cla, fhr, ds, site, workdir):
```
christinaholtNOAA, line 200: Nothing changed here. Just moved from create_graphics.py.

adb_graphics/figure_builders.py
```diff
@@ -0,0 +1,280 @@
...
+    plt.savefig(
+        png_path,
+        bbox_inches='tight',
+        dpi=cla.img_res,
+        format='png',
+        orientation='landscape',
+        )
+    plt.close()
+
+def set_figure(model_name, graphic_type, tile):
```
christinaholtNOAA, line 239: This is a new one that sets the size and aspect ratio for the plots. Square doesn't work well for the panel plots, and the same rectangle isn't great for all the subdomains.

adb_graphics/figures/maps.py
```diff
@@ -287,12 +284,28 @@ def _colorbar(self, cc, ax):
...
+        self._title()
+
+        # Create a pop-up to display the figure, if show=True
+        if show:
+            plt.tight_layout()
+            plt.show()
+
+        self.add_logo(self.map.ax)
+
+    def _draw_panel(self, wind_barbs=True): # pylint: disable=too-many-locals, too-many-branches
```
christinaholtNOAA, line 307: I split out the public `draw()` method here from the `_draw_panel()` method. Now `_draw_panel()` does all the things that are common to single and multi panel plots, and leaves the `draw()` method to do the extras (or not) like plotting colorbars/titles/logos. The addition of a subclass for multipanel plots lets us do a different kind of `draw()` method where we don't do all that by default with each panel.

adb_graphics/figures/maps.py
```diff
@@ -322,24 +332,14 @@ def draw(self, show=False): # pylint: disable=too-many-locals, too-many-branches
 
         # Add wind barbs, if requested
         add_wind = self.field.vspec.get('wind', False)
-        if add_wind:
+        if add_wind and wind_barbs:
             self._wind_barbs(add_wind)
```
christinaholtNOAA, line 336: Now we need the default_specs to have the `add_wind` and the argument to the `_draw_panel` method to also be true before we draw the winds. This allows us to turn off winds for all multipanel plots with one switch.

adb_graphics/figures/maps.py
```diff
@@ -322,24 +332,14 @@ def draw(self, show=False): # pylint: disable=too-many-locals, too-many-branches
...
-        # Finish with the title
-        self._title()
-
-        # Create a pop-up to display the figure, if show=True
-        if show:
-            plt.tight_layout()
-            plt.show()
-
-        self.add_logo(ax)
-
```
christinaholtNOAA, line 342: All this moved to the `draw()` method above.

adb_graphics/figures/maps.py
```diff
@@ -503,6 +502,16 @@ def _title(self):
         if contoured_units:
             contoured = f"{contoured} ({', '.join(contoured_units)}, contoured)"
 
+        return contoured
+
+    def _title(self):
```
christinaholtNOAA, line 507: Title is now implemented differently in the base class and subclass for multipanel plots.

adb_graphics/figures/maps.py
```diff
@@ -585,5 +595,168 @@ def _xy_mesh(self, field):
...
+        ax.text(0, 0.6,
+                f"{contoured}",
+                horizontalalignment='left',
+                verticalalignment='top',
+                fontsize=14,
+                transform=ax.transAxes,
+                )
+
+
+class MapFields():
```
christinaholtNOAA, line 706: This seemed like an okay-ish solution to remove a bunch of duplication in create_graphics for adding contours and hatches. Now we just send the spec, call them all "overlays" and let this more generalized code sort it out. We also can pass it the original "field" object and use it's common info so that we don't have to keep creating other objects with the same parameters.

adb_graphics/figures/maps.py
```diff
@@ -585,5 +595,168 @@ def _xy_mesh(self, field):
...
+        self.tile = kwargs.get('tile', 'full')
+
+    @property
+    def contours(self):
+        ''' Return the list of contour fieldData objects'''
+
+        # We won't plot contours on multipanel plots, or full global
+        # plots.
+        if self.map_type == 'enspanel':
+            return []
```
christinaholtNOAA, line 728: This allows us to turn off all contours on the ensemble plots. Similarly below, we can turn them off for global plots.

adb_graphics/utils.py
```diff
@@ -1,18 +1,59 @@
...
 import importlib as il
 from math import atan2, degrees
+from multiprocessing import Process
 import os
 import sys
 import time
 
 import numpy as np
 
+import yaml
```
christinaholtNOAA, line 19: These all just moved directly from `create_graphics.py`

create_graphics.py
```diff
@@ -1,5 +1,10 @@
 # pylint: disable=invalid-name
 '''
+                create_maps(cla,
+                            fhr=fhr,
+                            gribfiles=gribfiles,
+                            workdir=workdir,
+                            )
```
christinaholtNOAA, line 7: A copy/paste blunder. I'll fix that.

christinaholtNOAA (commented):
Oh! Also, I was able to get rid of the `libpng warning: iCCP: known incorrect sRGB profile error` by running "convert" on the logo file! That's the diff you see on that image.

christinaholtNOAA:
Examples! 
image
image
image


Note: The ensemble data and observation data are from separate valid times and are only meant to show design aspects!

adb_graphics/datahandler/gribdata.py
```diff
@@ -522,7 +525,7 @@ def corners(self) -> list:
         '''
 
         lat, lon = self.latlons()
-        if self.model == 'global':
+        if self.model in ['global', 'obs']:
```
cshartsough, line 528: corner values are in a different order for global/obs vs the rest?

adb_graphics/figures/maps.py
cshartsough: I like this way of doing it.  Nice.

cshartsough (approved):
This looks great.  I think I only had one editing comment, which is just to double check one line.  Otherwise I think it's ready.

Just want to note:  this is a HUGE amount of work and very impressive in how the changes improve the workflow and handling of the many (and growing) varieties of maps.  Not saying that with any surprise, just feel like it needs to be pointed out how invaluable Christina's work is for this project.

adb_graphics/datahandler/gribdata.py
christinaholtNOAA: Yes. They are stored and a 1-D array of lat/lon instead of meshed together in 2D arrays. Every grid box has the same lat along a lon, and vice versa.

christinaholtNOAA (commented):
I tested on Jet with HRRR single maps, and needed one tiny change, but things are running smoothly (as expected) otherwise.

Brian-Jamison (approved):
Wow, this is a lot of work!  It's hard to wrap my head around all of the changes, but they look good.  I'm not quite sure about modifying the aspect ratio on some of the plots, but if they don't look too "stretched" it should be ok.  Are you using a HRRR or RRFS analysis for the obs panel in your examples?
Lots of great work, here!  Thanks!!

christinaholtNOAA:
@Brian-Jamison The observations are the composite reflectivity observations. They come to us in grib2 format, so that was a mostly straightforward addition. It requires reading in an additional file, but nothing too major.

PR #158, opened 2022-04-22
Write CSV files of vertical profiles when running SkewTs

christinaholtNOAA:
This feature was contributed by @terraladwig in response to the need to deliver the vertical profiles in CSV format to downstream stakeholders in the Spring Experiment. These files will now be written by default when deployed, and zipped separately from the png files. 

The need is ultimately for RRFS and RTMA, but it shouldn't hurt to have this running for HRRR and RAP, too. It didn't take too many additional resources (compute or disk) and should be cleaned from disk with the existing cleaning processes.

A quick note...I have applied these changes as a patch from work I was doing in a separate branch and the git history on the current branch does not accurately reflect the contributors. My apologies for losing that commit history. If that is a problem, please let me know and I will attempt to preserve that history more accurately.

### DIFF
--- a/adb_graphics/figure_builders.py
+++ b/adb_graphics/figure_builders.py
@@ -234,6 +234,15 @@ def parallel_skewt(cla, fhr, ds, site, workdir):
         format='png',
         orientation='landscape',
         )
+
+    start_time = cla.start_time.strftime('%Y%m%d%H')
+    csvfile = f"{skew.site_code}.{skew.site_num}.skewt.{start_time}_f{fhr:03d}.csv"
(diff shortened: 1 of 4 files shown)


cshartsough (approved):
Looks good. Will be very useful.

Brian-Jamison (approved):
LGTM!  I did something similar for Jordan, recently, that makes profiles of data for RAP-Chem variables.  I did it using NCL, though.  This is more versatile, so maybe in the future, I'll change that process to use this, instead.  Great job!

terraladwig:
Thanks for merging the changes!

PR #161, opened 2022-04-27
Initial changes for scatter plot capability.

Brian-Jamison:
Please note that this PR is intended for discussion only.  It will not pass pylint or pytest, and won't be merged.

I'm pushing this PR, complete with all my dirty laundry laying around.  I was finally successful in getting the plot to come out with adjusted sizes for the dots.  I realized that the values themselves were being used to scale the size, with the help of a default algorithm.  These are adjusted by adjusting the values in the array in the _plot_field function.  For some reason, I can only get the scatter function to accept named Matplotlib colors.  I matched them as closely as I could to the actual colormap.

Here's an example:
image

### DIFF
--- a/adb_graphics/default_specs.yml
+++ b/adb_graphics/default_specs.yml
@@ -806,8 +806,18 @@ mfrp: # Fire radiative power
   sfc:
     clevs: [0, 10, 25, 50, 100, 250]
     colors: fire_power_colors
+    # cmap: fire_power_colors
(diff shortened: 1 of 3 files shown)


christinaholtNOAA:
@Brian-Jamison Should we close this one since you've opened #171?

christinaholtNOAA:
Reopened as PR #171

PR #162, opened 2022-04-29
Fix enspanel obs, and skewt csv zips.

christinaholtNOAA:
Both of my last couple of PRs need some minor fixes to the logic when these items shouldn't be done.


Brian-Jamison (approved):
LGTM!

PR #163, opened 2022-04-29
use negative range and reversed colors for min hlcy

cshartsough:
Changes to defaults_specs.yml and specs.py to use negative range and a reversed color table for minimum helicity relative to max helicity.

Minimum helicity values are generally small and don't show in the field, however, David D. thought the range worked as a symmetry between max and min helicity graphics.  May need to revisit the range later if others want to see more detail.

Passed linter.

### DIFF
--- a/adb_graphics/default_specs.yml
+++ b/adb_graphics/default_specs.yml
@@ -623,33 +623,23 @@ hlcy: # Helicity
     ncl_name: UPHL_P0_2L103_{grid}
     split: True
     title: 2-5km Updraft Helicity
-  mn02: &hlcy_mn02 # Hourly minimum of updraft helicity over 0-2 km layer
-    <<: *hlcy
-    clevs: [-300,-275,-250,-225,-200,-175,-150,-125,-100,-75,-50,-25]
+  mn02: &hlcy_mn # Hourly minimum of updraft helicity over 0-2 km layer
+    clevs: !!python/object/apply:numpy.arange [-300, -24, 25]
+    cmap: gist_ncar
     colors: rainbow12_reverse
     ncl_name: VAR_0_7_200_P8_2L103_{grid}_min1h
     split: True
+    ticks: 25
     title: 0-2km Min Updraft Helicity (over prv hr)
+    unit: $m^2 / s^2$
   mn03: &hlcy_mn03 # Hourly minimum of updraft helicity over 0-3 km layer
-    <<: *hlcy
-    clevs: [-300,-275,-250,-225,-200,-175,-150,-125,-100,-75,-50,-25]
-    colors: rainbow12_reverse
-    ncl_name: VAR_0_7_200_P8_2L103_{grid}_min1h
-    split: True
+    <<: *hlcy_mn
     title: 0-3km Min Updraft Helicity (over prv hr)
   mn16: &hlcy_mn16 # Hourly minimum of updraft helicity over 1-6 km layer
-    <<: *hlcy
-    clevs: [-300,-275,-250,-225,-200,-175,-150,-125,-100,-75,-50,-25]
-    colors: rainbow12_reverse
-    ncl_name: VAR_0_7_200_P8_2L103_{grid}_min1h
-    split: True
+    <<: *hlcy_mn
     title: 1-6km Min Updraft Helicity (over prv hr)
   mn25: &hlcy_mn25 # Hourly minimum of updraft helicity over 2-5 km layer
-    <<: *hlcy
-    clevs: [-300,-275,-250,-225,-200,-175,-150,-125,-100,-75,-50,-25]
-    colors: rainbow12_reverse
-    split: True
-    ncl_name: VAR_0_7_200_P8_2L103_{grid}_min1h
+    <<: *hlcy_mn
     title: 2-5km Min Updraft Helicity (over prv hr)
   mx02: &hlcy_mx02 # Hourly maximum of updraft helicity over 0-2 km layer
     <<: *hlcy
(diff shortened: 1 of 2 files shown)


adb_graphics/default_specs.yml
```diff
@@ -625,27 +625,29 @@ hlcy: # Helicity
     title: 2-5km Updraft Helicity
   mn02: &hlcy_mn02 # Hourly minimum of updraft helicity over 0-2 km layer
     <<: *hlcy
-    clevs: !!python/object/apply:numpy.arange [12.5, 150.5, 12.5]
+    clevs: [-300,-275,-250,-225,-200,-175,-150,-125,-100,-75,-50,-25]
```
christinaholtNOAA, line 628: Can we use a range here? And maybe an anchor so that the ones below can just reference this setting and all show up the same? Just to reduce duplication.

adb_graphics/specs.py
```diff
@@ -298,6 +298,16 @@ def rainbow12_colors(self) -> np.ndarray:
...
 
+    @property
+    def rainbow12_reverse(self) -> np.ndarray:
+
+        ''' Default color map for min helicity '''
+
+        grays = cm.get_cmap('Greys', 2)([0])
+        ncar = cm.get_cmap(self.vspec.get('cmap'), 128) \
+                          ([120, 100, 90, 85, 80, 70, 60, 50, 25, 20, 18, 15])
+        return np.concatenate((ncar, grays))
```
christinaholtNOAA, line 309: One thing that could help here is:

```
def rainbow12_reverse(self) -> np.ndarray:
    return np.flip(rainbow12_colors)
```

christinaholtNOAA (changes requested):
@cshartsough Thanks for tackling this! I have a couple of suggestions below that could reduce some duplication of information.

adb_graphics/default_specs.yml
cshartsough: These fields are already "double anchored".  I might need some advice on how to get a third level of anchoring in, or to modify the existing anchors.

Brian-Jamison:
Good work, Craig!  I agree with Christina's suggestions.
In your explanation, you mention that min helicity values are "generally small", but when it was explained to me, the negative (i.e. "min") values can be as large as the positive (i.e. "max") values.  Tornadoes can rotate clockwise in the Northern Hemisphere, though they are rare.  I like keeping the scale as it is.

adb_graphics/specs.py
cshartsough: Having trouble with the flip command, I think because the `rainbow12_colors` array isn't a 1-D set of indices, it's a set of 4-value tuples(?).  So in doesn't just flip the order of colors, it flips the 4 values in each color index:
```
array([[1.        , 1.        , 1.        , 1.        ],
       [0.        , 0.17242769, 1.        , 1.        ],
       [0.        , 0.50600805, 1.        , 1.        ],
       [0.        , 0.72518476, 1.        , 1.        ],
       [0.        , 0.91691389, 1.        , 1.        ],
       [0.44128587, 0.88987998, 0.        , 1.        ],
       [0.72303033, 1.        , 0.2259988 , 1.        ],
       [1.        , 0.92704406, 0.        , 1.        ],
       [1.        , 0.73284702, 0.05330818, 1.        ],
       [1.        , 0.41889133, 0.01952186, 1.        ],
       [1.        , 0.17326918, 0.        , 1.        ],
       [1.        , 0.        , 0.88868695, 1.        ],
       [0.95686887, 0.71717346, 0.96259683, 1.        ]])
```
flips to:
```
array([[1.        , 0.96259683, 0.71717346, 0.95686887],
       [1.        , 0.88868695, 0.        , 1.        ],
       [1.        , 0.        , 0.17326918, 1.        ],
       [1.        , 0.01952186, 0.41889133, 1.        ],
       [1.        , 0.05330818, 0.73284702, 1.        ],
       [1.        , 0.        , 0.92704406, 1.        ],
       [1.        , 0.2259988 , 1.        , 0.72303033],
       [1.        , 0.        , 0.88987998, 0.44128587],
       [1.        , 1.        , 0.91691389, 0.        ],
       [1.        , 1.        , 0.72518476, 0.        ],
       [1.        , 1.        , 0.50600805, 0.        ],
       [1.        , 1.        , 0.17242769, 0.        ],
       [1.        , 1.        , 1.        , 1.        ]])
```

cshartsough: nevermind, figured it out

Brian-Jamison (approved):
LGTM!

adb_graphics/default_specs.yml
```diff
@@ -623,31 +623,23 @@ hlcy: # Helicity
...
-    ncl_name: VAR_0_7_200_P8_2L103_{grid}_min1h
-    split: True
+    <<: *hlcy_mn
     title: 1-6km Min Updraft Helicity (over prv hr)
   mn25: &hlcy_mn25 # Hourly minimum of updraft helicity over 2-5 km layer
-    <<: *hlcy
-    clevs: !!python/object/apply:numpy.arange [25, 301, 25]
-    split: True
-    ncl_name: VAR_0_7_200_P8_2L103_{grid}_min1h
+    <<: *hlcy_mn
```
christinaholtNOAA, line 642: This cleaned up really nicely! Thanks!

adb_graphics/specs.py
```diff
@@ -298,6 +298,13 @@ def rainbow12_colors(self) -> np.ndarray:
                           ([15, 18, 20, 25, 50, 60, 70, 80, 85, 90, 100, 120])
         return np.concatenate((grays, ncar))
 
+    @property
+    def rainbow12_reverse(self) -> np.ndarray:
+
+        ''' Default color map for min helicity '''
+
+        return np.flip(self.rainbow12_colors, 0)
```
christinaholtNOAA, line 306: Sorry I didn't get back to you on this question in a timely manner. You found it! Thanks!

christinaholtNOAA (approved):
Looks great! Thanks for taking the time to make those changes.

PR #164, opened 2022-05-06
Update zip instead of append.

christinaholtNOAA:
We will need this functionality to be able to deliver updated ens panel plots that have observations. 

I attempted to do this with python-native utilities, but that proved quite expensive, so I switched to a subprocess and call of the linux-native zip utility.

### DIFF
--- a/adb_graphics/utils.py
+++ b/adb_graphics/utils.py
@@ -11,9 +11,9 @@ import importlib as il
 from math import atan2, degrees
 from multiprocessing import Process
 import os
+import subprocess
 import sys
 import time
-import zipfile
 
 import numpy as np
 import yaml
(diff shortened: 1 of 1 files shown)


adb_graphics/utils.py
```diff
@@ -246,34 +249,6 @@ def uniq_wgrib2_list(inlist):
...
-            file_tmpl = f'*_{tile}_*{fhr:02d}.png'
-        product_files = glob.glob(os.path.join(workdir, file_tmpl))
-        if product_files:
-            zip_proc = Process(group=None,
-                               target=create_zip,
-                               args=(product_files, zipf),
-                               )
-            zip_proc.start()
-            zip_proc.join()
-
```
christinaholtNOAA, line 276: Just needed to move this down to be able to use the timer decorator, and because it was out of alphabetical order. I didn't make a change to the contents.

adb_graphics/utils.py
```diff
@@ -325,11 +300,40 @@ def wrapper_timer(*args, **kwargs):
         value = func(*args, **kwargs)
         toc = time.perf_counter()
         elapsed_time = toc - tic
-        print(f"Elapsed time: {elapsed_time:0.4f} seconds")
+        print(f"{func.__name__} Elapsed time: {elapsed_time:0.4f} seconds")
```
christinaholtNOAA, line 303: It's helpful to see which function we're timing in the log.

cshartsough (approved):
good work, it'll be very helpful to the ensemble team.

Brian-Jamison (approved):
Looks great!

PR #166, opened 2022-08-02
Use a new equation for Fire Weather Index.

christinaholtNOAA:
@EricJames-NOAA requested a new formulation of the Fire Weather Index via email. It is implemented here.


cshartsough (approved):
Looks like the requested change is properly implemented, hopefully it works well.

PR #167, opened 2022-08-02
Fix environment build problem.

christinaholtNOAA:
Tried updating the environment.yml file for the conda installation for GHA in a jillion different ways.

Ended up being an issue with a change to the default value of channel priority.


Brian-Jamison (approved):
LGTM!

cshartsough (approved):
Thanks for tackling this.

PR #168, opened 2022-08-11
Enable running enspost maps for RRFSE.

christinaholtNOAA:
The changes proposed here were tested while RRFSE was running in real time. 

I have linted and run pytest on the code locally with both passing. The GHA workflows were still broken earlier this week when I attempted to run them again.


Brian-Jamison (approved):
LGTM!

PR #169, opened 2022-08-31
Changed default_specs.yml for GLC0 grid to {grid} for ceiling field designation.

Brian-Jamison:
The specific reference of GLC0 for ceiling caused problems in HRRR-AK, since it is a polar stereographic grid and requires that variables end in GST0.  Using the {grid} designation, this problem has been tested and resolved.


cshartsough (approved):
I'm sure that was an old bit of my code, from before we generalized the grids.  Thanks for fixing it.

christinaholtNOAA:
@Brian-Jamison I went ahead and merged main in here so we can see the tests (expecting all to pass). This is testing that fix, for the most part.

christinaholtNOAA (approved):

PR #171, opened 2022-09-12
mfrp_scatter_plot

Brian-Jamison:
Contours of Mean Fire Radiative Power are difficult to see, especially in larger domains.  These changes provide the ability to plot values as a scatter plot, mimicking the NCL plots for MFRP.

image
image
image
image


adb_graphics/figures/maps.py
```diff
@@ -394,6 +428,11 @@ def _draw_field(self, ax, field, func, **kwargs):
         x, y = self._xy_mesh(field)
         vals = field.values()
 
+        # Scatter plot dots are sized by value. Doing this here alters the size
+        # without altering the colors.
+        if self.plot_scatter:
+            vals = np.full_like(vals, np.log10(vals) * 20, dtype='float')
+
```
christinaholtNOAA, line 435: It would be best to keep all this logic contained together with this special plot.

adb_graphics/figures/maps.py
```diff
@@ -372,6 +378,34 @@ def _draw_contours(self, ax, not_labeled):
...
                             {self.field.level}')
 
+    def _draw_scatter(self, ax):
+
+        ''' Plot dots at locations on the map that meet a threshold. '''
+
+        field = self.field
+        levels = self.field.clevs
+        alpha_val = 0.75 # set opacity
+        vals = self.field.values()
```
christinaholtNOAA, line 388: It might be good to put the size logic here. Maybe something like:

```
sizes = np.full_like(self.field.values(), np.log10(vals) * 20, dtype='float')
```

Using `full_like` will not cause problems when you change the values. The data is not linked back to reference those saved in `self.field`.

[This example](https://www.pythonprogramming.in/adjust-marker-sizes-and-colors-in-scatter-plot.html) shows that you might be able to provide the size and color as `s` and `c` args in the call to `scatter`.

adb_graphics/figures/maps.py
```diff
@@ -372,6 +378,34 @@ def _draw_contours(self, ax, not_labeled):
...
+        colors = ['white', 'lightskyblue', 'darkblue', 'green', 'darkorange', \
+                  'indianred', 'firebrick']
+
+        ci = copy.copy(vals)
+        ci = np.full_like(ci, colors[0], dtype='object')
+        ci = np.where((vals > levels[0]) & (vals <= levels[1]), colors[1], ci)
+        ci = np.where((vals > levels[1]) & (vals <= levels[2]), colors[2], ci)
+        ci = np.where((vals > levels[2]) & (vals <= levels[3]), colors[3], ci)
+        ci = np.where((vals > levels[3]) & (vals <= levels[4]), colors[4], ci)
+        ci = np.where((vals > levels[4]) & (vals <= levels[5]), colors[5], ci)
```
christinaholtNOAA, line 398: A fun python syntax trick:

```
levels[3] < vals <= levels[4]
```

A loop might also be a good thing here.
```
ci = np.full_like(ci, colors[0], dtype='object')
for i, color in enumerate(colors):
    ci = np.where(levels[i] < vals < levels[i+1], color[i+1], ci)
```

christinaholtNOAA (commented):
These graphics look very nice, @Brian-Jamison!

I have some suggestions below. Please let me know if they got jumbled and unclear.

adb_graphics/figures/maps.py
```diff
@@ -341,6 +343,10 @@ def _draw_panel(self, wind_barbs=True): # pylint: disable=too-many-locals, too-m
         if annotate and 'global' not in self.map.model: # too dense in global
             self._draw_field_values(ax)
 
+        # Add scatter plot, if requested
+        if self.plot_scatter:
+            self._draw_scatter(ax)
+
```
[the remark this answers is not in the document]
Brian-Jamison: Leaving this as it is may allow for contours or hatches to be plotted along with the scatter plot markers, which has a fair possibility of being requested.  I could imagine smoke contours on the plot, but I haven't tested this.

adb_graphics/figures/maps.py
Brian-Jamison: Matplotlib's scatter plot method is limited.  To do the scatter plot on a map background requires the field, which has the lat,lon (x,y) info.  The scatter plot method will use the values associated with the field as the size field, even if the sizes are provided to the method in a 1-dimensional format (also required by the scatter plot method) the same size as the colors array.  If I leave "field=field" and add the sizes field, it will crash on having two instances of the size array.  If I try to set "field=sizes", it will crash because it doesn't have the lat,lon information.  This is why the field needs to be manipulated after the colors have been set using the original values, and then the values can be manipulated to adjust the sizes.

adb_graphics/figures/maps.py
Brian-Jamison: I know and I've tried keeping this statement within the _draw_scatter function.  If I put the statement there, and comment out the one in _draw_field, I get this error:

  File "/mnt/lfs1/BMC/wrfruc/jamison/adb_graphics/adb_graphics/figures/maps.py", line 640, in _xy_mesh
    lat, lon = field.latlons()
AttributeError: 'numpy.ndarray' object has no attribute 'latlons'

adb_graphics/figures/maps.py
```diff
@@ -372,6 +378,34 @@ def _draw_contours(self, ax, not_labeled):
...
+        ''' Plot dots at locations on the map that meet a threshold. '''
+
+        field = self.field
+        levels = self.field.clevs
+        alpha_val = 0.75 # set opacity
+        vals = self.field.values()
+        colors = ['white', 'lightskyblue', 'darkblue', 'green', 'darkorange', \
+                  'indianred', 'firebrick']
+
+        ci = copy.copy(vals)
```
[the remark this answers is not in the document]
Brian-Jamison: Done and done.  Thanks!

adb_graphics/figures/maps.py
Brian-Jamison: The loop is a good idea, and I've tried the loop you suggested.  It worked up until the last iteration, and then crashes because of referencing levels[i+1] which is outside of the levels array.  I had to change it to account for this.  I changed the loop to use levels instead of colors, and added an exception for the last iteration.  The enumerate part caused a problem in pylint, since I never used the "level" variable in "for level in enumerate(levels)".  I changed it to just use the length of the levels array.  The "fun python trick" didn't work, as it complained that I needed to use vals.any or vals.all instead of just vals, and the results weren't correct in either case.

adb_graphics/figures/maps.py
```diff
@@ -372,6 +378,34 @@ def _draw_contours(self, ax, not_labeled):
...
+        ci = np.full_like(ci, colors[0], dtype='object')
+        ci = np.where((vals > levels[0]) & (vals <= levels[1]), colors[1], ci)
+        ci = np.where((vals > levels[1]) & (vals <= levels[2]), colors[2], ci)
+        ci = np.where((vals > levels[2]) & (vals <= levels[3]), colors[3], ci)
+        ci = np.where((vals > levels[3]) & (vals <= levels[4]), colors[4], ci)
+        ci = np.where((vals > levels[4]) & (vals <= levels[5]), colors[5], ci)
+        ci = np.where(vals > levels[5], colors[6], ci)
```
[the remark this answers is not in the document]
Brian-Jamison: I moved the colors array definition to specs.py, and the levels are designated in default_specs.py.  I left the rest here, since the field is needed to assign the colors.

adb_graphics/figure_builders.py
```diff
@@ -142,7 +142,8 @@ def parallel_maps(cla, fhr, ds, level, model, spec, variable, workdir,
             map_=m,
             member=mem,
             model_name=cla.model_name,
-            last_panel=last_panel
+            last_panel=last_panel,
+            plot_scatter=spec.get('plot_scatter', False),
```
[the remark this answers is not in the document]
Brian-Jamison: I've changed it, thanks!

adb_graphics/figures/maps.py
christinaholtNOAA: With my recent PR for handling diff plots, you should be able to set the `data` attribute of the field object. So, in `_draw_scatter` you can set the values you want to plot like this:

```
field.data = np.log10(field.values()) * 20
```

Then you should have the values you want to plot and the lat/lon attributes you need from the field object.

christinaholtNOAA (commented):
@Brian-Jamison I'm really sorry for not getting back to this review for so long! 

It is looking really good. Just a couple of things below.

adb_graphics/figures/maps.py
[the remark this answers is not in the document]
Brian-Jamison: I believe so, but they are plotted over with the dots, which are bigger, and are the same color.

adb_graphics/figures/maps.py
Brian-Jamison: Thank you for this, it worked!  Now all of the logic is contained within _draw_scatter.

christinaholtNOAA (approved):
This looks great @Brian-Jamison! Thanks!

PR #173, opened 2022-10-03
Nyc boston wfip3 enhanced

Brian-Jamison:
Dave Turner and Joe Olson requested some changes to the NYC-BOS HRRR subdomain.  These include: shifting the boundaries to the south and east, and adding marked WFIP3 locations.  Since there are only 2 locations that are both located on off-coast islands, and those islands have airports, the new locations were added to the aiports data files.

The following image illustrates these two WFIP3 locations (plotted here without the other airports).
image


christinaholtNOAA (approved):
LGTM!

PR #174, opened 2022-10-05
Vis colors update

cshartsough:
This updates the visibility colors, as originally suggested by Stan Benjamin.  It add a lower-resolution color scheme in grays to the larger visibility range, to provide some additional detail.

The changes required a new util.py method, join_ranges, which concatenates a list of ranges from the default_specs.yml file.

Passed pylint test.

Sample original and new plots below (different date/times)
image
image

### DIFF
--- a/adb_graphics/specs.py
+++ b/adb_graphics/specs.py
@@ -396,6 +396,8 @@ class VarSpec(abc.ABC):
         IFR (Instrument Flight Rules) -- 1 mile to less than 3 miles
         MVFR (Marginal Visual Flight Rules) -- 3 to 5 miles
         VFR (Visual Flight Rules) -- greater than 5 miles
+
+        the gray range is arbitrary compared to the official flight levels
         '''
 
         lifr = cm.get_cmap('RdPu_r', 20)(range(0, 11))
--- a/adb_graphics/specs.py
+++ b/adb_graphics/specs.py
@@ -402,10 +402,10 @@ class VarSpec(abc.ABC):
         ifr = cm.get_cmap('autumn', 30)(range(0, 30))
         mvfr = cm.get_cmap('Blues', 20)(range(10, 20))
         vfr1 = cm.get_cmap('YlGn_r', 60)(range(0, 50))
-        vfr2 = cm.get_cmap('Greys', 20)(np.full(100, 9))
-        hi01 = cm.get_cmap('Greys', 20)(np.full(100, 6))
-        hi02 = cm.get_cmap('Greys', 20)(np.full(200, 3))
-        hi03 = cm.get_cmap('Greys', 20)(np.full(1, 0))
+        vfr2 = cm.get_cmap('Greys', 25)(np.full(10, 9))
+        hi01 = cm.get_cmap('Greys', 25)(np.full(10, 6))
+        hi02 = cm.get_cmap('Greys', 25)(np.full(20, 3))
+        hi03 = cm.get_cmap('Greys', 25)(np.full(1, 0))
 
         return np.concatenate((lifr, ifr, mvfr, vfr1, vfr2, hi01, hi02, hi03))
 
--- a/adb_graphics/specs.py
+++ b/adb_graphics/specs.py
@@ -398,7 +398,7 @@ class VarSpec(abc.ABC):
         VFR (Visual Flight Rules) -- greater than 5 miles
         '''
 
-        lifr = cm.get_cmap('RdPu_r', 20)(range(0, 10))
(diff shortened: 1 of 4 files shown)


adb_graphics/utils.py
```diff
@@ -115,6 +115,34 @@ def get_func(val: str):
...
+
+    e.g.: in default_specs.yml, clevs for visibility can be assigned as
+
+        clevs: !join_ranges [[0, 10, 0.1], [10, 51, 1.0]]
+
+    The join_ranges method concatenates these ranges into a single array of levels.
+    This can be useful for plots where one part of the color ranges requires higher
+    resolution than the rest, while keeping the colorbar from looking squished.
+
+    Note that a "yaml.add_constructor" is required, as shown after the method.
```
christinaholtNOAA, line 132: Would you mind putting a pointer (just text that says something like "go here to read more") in default specs under clevs to this file and method's docs?

adb_graphics/utils.py
```diff
@@ -115,6 +115,34 @@ def get_func(val: str):
...
+    for seq_node in node.value:
+        range_args = []
+        for scalar_node in seq_node.value:
+            range_args.append(float(scalar_node.value))
+
+        list_.append(np.arange(*range_args))
+
+    return np.concatenate(list_, axis=0)
+
+yaml.add_constructor("!join_ranges", join_ranges, Loader=yaml.Loader)
```
christinaholtNOAA, line 145: I recall you mentioning in person that there are issues with using the SafeLoader here. It'd be nice if we could use the SafeLoader like you do in the tests. What was it complaining about?

tests/test_common.py
```diff
@@ -113,6 +114,30 @@ def test_utils():
...
+    expected = np.concatenate((np.arange(0, 15, 0.1),
+                               np.arange(20, 61, 20)), axis=0)
+    expected2 = np.arange(0, 15, 0.1)
+    expected3 = np.concatenate((np.arange(0, 15, 0.1),
+                                np.arange(20, 40, 10),
+                                np.arange(40, 61, 20)), axis=0)
+
+    assert np.array_equal(expected, cfg['foo'])
+    assert np.array_equal(expected2, cfg['foo2'])
+    assert np.array_equal(expected3, cfg['foo3'])
```
christinaholtNOAA, line 138: Nice tests! :)

adb_graphics/specs.py
```diff
@@ -402,9 +402,12 @@ def vis_colors(self) -> np.ndarray:
...
         mvfr = cm.get_cmap('Blues', 20)(range(10, 20))
         vfr1 = cm.get_cmap('YlGn_r', 60)(range(0, 50))
-        vfr2 = cm.get_cmap('Reds', 15)(np.full(51, 1))
+        vfr2 = cm.get_cmap('Greys', 25)(np.full(10, 9))
+        hi01 = cm.get_cmap('Greys', 25)(np.full(10, 6))
+        hi02 = cm.get_cmap('Greys', 25)(np.full(20, 3))
+        hi03 = cm.get_cmap('Greys', 25)(np.full(1, 0))
 
-        return np.concatenate((lifr, ifr, mvfr, vfr1, vfr2))
+        return np.concatenate((lifr, ifr, mvfr, vfr1, vfr2, hi01, hi02, hi03))
```
christinaholtNOAA, line 410: Did Stan provide any categories for these additional gray ranges? It might be useful to add those to the doc block above. If he didn't, perhaps adding a line to indicate that an arbitrary gray scale covers values with greater than 5 mi (?) of visibility. "Arbitrary" would be important here since the other categories have solid rules in the aviation world.

christinaholtNOAA (approved):
The code changes look good! I have a couple of requests for additional documentation below, and a couple of questions. I'm good with it after those, though.

adb_graphics/utils.py
cshartsough: SafeLoader doesn't seem compatible with our numpy contructors, using Loader here.

adb_graphics/utils.py
cshartsough: added this to the clevs for visibility in default_specs:

"see the description provided in adb_graphics/utils.py, for the join_ranges method."

adb_graphics/specs.py
cshartsough: added this to the vis_colors method in specs.py:

"the gray range is arbitrary compared to the official flight levels"

Brian-Jamison (approved):
LGTM!  Nice job!

PR #176, opened 2022-10-07
Global product additions

Brian-Jamison:
When implementing the global graphics, some products were noted to be missing.  Most of these were just omitted from the image lists, but a couple of them needed some special attention.  The mean 850mb-500mb RH uses levels at every 25mb between those ranges, but the global models don't have all of the levels at this interval, so these were specified as "global_levels".  Also, the 500mb height plot is a filled plot for global models, whereas only contours of 500mb height appear on the 500mb temperature and vorticity plots for other models.  Plotting height with the current setup had a multiple repeating color table in the color bar, and was unusable.  Here, it was changed to something more appropriate.

Some examples of the new products are shown below.

image
image
image
image


adb_graphics/datahandler/gribdata.py
```diff
@@ -112,17 +112,24 @@ def field_diff(self, values, variable2, level2, **kwargs):
...
+        if 'global' in self.model:
+            for level in global_levels:
+                val_lev = self.values(name=variable, level=level)
+                fsum = fsum + val_lev
+                val_lev.close()
+        else:
+            for level in levels:
+                val_lev = self.values(name=variable, level=level)
+                fsum = fsum + val_lev
+                val_lev.close()
```
christinaholtNOAA, line 132: So we don't have to duplicate the algorithm here, it would be cleaner to choose the levels you want to plot with logic, then do the heavy lifting using that variable. Something like this:

```
chosen_levels = global_levels if 'global' in self.model else levels
for level in chosen_levels:
    val_lev = ...
    fsum += val_lev
    val_lev.close()
```

adb_graphics/datahandler/gribdata.py
```diff
@@ -526,7 +533,7 @@ def corners(self) -> list:
         '''
 
         lat, lon = self.latlons()
-        if self.model in ['global', 'obs']:
+        if self.model in ['global', 'obs'] and self.model != 'globalAK':
```
christinaholtNOAA, line 536: The original logic only looks for `self.model == 'global'` and `self.model == 'obs'`, so I'm not sure this additional logic is needed for `globalAK`.

adb_graphics/default_specs.yml
```diff
@@ -1547,6 +1562,42 @@ wspeed: # Wind Speed
         field2: VGRD_P0_L103_{grid}
     unit: kt
     wind: True
+  5mb:
```
christinaholtNOAA, line 1565: Perhaps it would be helpful to add an anchor here that can be referenced in the other new levels below? This would reduce duplication.

```
5mb: &contoured_ua_wspeed
...
10mb:
  <<: *contoured_ua_wspeed
  title: 10mb Wind
...
```

christinaholtNOAA (commented):
This looks good. Just a few comments below.

adb_graphics/datahandler/gribdata.py
Brian-Jamison: Thanks, I suspected there was a better way to do this.

adb_graphics/datahandler/gribdata.py
Brian-Jamison: I'm not sure why I thought this was needed.  I've removed it.

adb_graphics/default_specs.yml
Brian-Jamison: I'm not sure why the changes aren't showing up, but I made this change, though I used "shaded_ua_wspeed" instead of "contoured_ua_wspeed".

christinaholtNOAA (approved):
LGTM! Thanks @Brian-Jamison!!

PR #180, opened 2022-10-25
add RRFS FIP icing products

cshartsough:
This update adds new icing products to the main python graphics repository.  The products include Icing Probability (icprb), Icing Severity (icsev), and Supercooled Large Droplet Icing (sipd).   There are 60 vertical levels, in 500' increments up to 30,000ft.  Awaiting confirmation on which levels to plot by default, and will adjust _default_specs.yml_ accordingly.

This update also includes a new method in gribdata.py to compute the column_max for any field.  Testing should already be included in the current test_common.py steps.

Passed pylint and pytest.

sample plots:
Presentation2

### DIFF
--- a/adb_graphics/datahandler/gribdata.py
+++ b/adb_graphics/datahandler/gribdata.py
@@ -100,6 +100,17 @@ class UPPData(specs.VarSpec):
 
         return self._get_field(self.ncl_name(self.vspec))
 
+    def field_column_max(self, values, variable, level, **kwargs):
+
+        # pylint: disable=unused-argument
+
+        ''' Returns the column max of the values. '''
+
+        vals = self.values(name=variable, level=level, one_lev=False)
+        maxvals = vals.max(axis=0)
+
+        return maxvals
+
     def field_diff(self, values, variable2, level2, **kwargs):
 
         # pylint: disable=unused-argument
@@ -111,17 +122,6 @@ class UPPData(specs.VarSpec):
         value2.close()
 
         return diff
-
-    def field_column_max(self, values, variable, **kwargs):
-
-        # pylint: disable=unused-argument
-
-        ''' Returns the column max of the values. '''
-
-        vals = self.values(name=variable, level='1000ft', one_lev=False)
-        maxvals = vals.max(axis=0)
-
-        return maxvals
 
     def field_mean(self, values, variable, levels, **kwargs):
 
--- a/adb_graphics/datahandler/gribdata.py
+++ b/adb_graphics/datahandler/gribdata.py
@@ -1,4 +1,4 @@
-# pylint: disable=invalid-name,too-few-public-methods
+# pylint: disable=invalid-name,too-few-public-methods, too-many-lines
 
 '''
 Classes that handle the specifics of grib files from UPP.
@@ -113,6 +113,8 @@ class UPPData(specs.VarSpec):
         return diff
 
     def field_column_max(self, values, variable, **kwargs):
+
+        # pylint: disable=unused-argument
 
         ''' Returns the column max of the values. '''
 
--- a/adb_graphics/datahandler/gribdata.py
+++ b/adb_graphics/datahandler/gribdata.py
@@ -112,6 +112,15 @@ class UPPData(specs.VarSpec):
 
         return diff
 
+    def field_column_max(self, variable, **kwargs):
+
+        ''' Returns the column max of the values. '''
+
+        icprb = self.values(name='sipd', level='1000ft', one_lev=False)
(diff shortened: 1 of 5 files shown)


adb_graphics/datahandler/gribdata.py
```diff
@@ -112,6 +112,17 @@ def field_diff(self, values, variable2, level2, **kwargs):
...
 
+    def field_column_max(self, values, variable, **kwargs):
+
+        # pylint: disable=unused-argument
+
+        ''' Returns the column max of the values. '''
+
+        vals = self.values(name=variable, level='1000ft', one_lev=False)
+        maxvals = vals.max(axis=0)
+
```
Brian-Jamison, line 123: Is the selection of level='1000ft' arbitrary, since you set one_lev=False, or is there a particular reason to select this level?
I assume axis=0 is the vertical axis?

adb_graphics/datahandler/gribdata.py
```diff
@@ -112,6 +112,17 @@ def field_diff(self, values, variable2, level2, **kwargs):
 
         return diff
 
+    def field_column_max(self, values, variable, **kwargs):
+
+        # pylint: disable=unused-argument
+
+        ''' Returns the column max of the values. '''
+
+        vals = self.values(name=variable, level='1000ft', one_lev=False)
```
christinaholtNOAA, line 121: I'm curious if the level arg here is necessary. I was thinking that if it's not set, it might be useful for variables that don't have a 1000ft level set.

adb_graphics/default_specs.yml
```diff
@@ -740,6 +740,56 @@ icmr: # Ice Water Mixing Ratio
...
+        variable: icprb
+icsev: # Icing Severity
+  500ft: &icsev
+    clevs: [0.0, 0.5, 1.0, 2.0, 3.0]
+    cmap: Blues
+    colors: icsev_colors
+    ncl_name: ICSEV_P0_L102_{grid}
+    ticks: 0
+    vertical_index: 0   # grib2 vertical levels are in meters,
+                        # so using index here to avoid errors from conversions
```
Brian-Jamison, line 777: If the number of vertical levels changes in the future, this will have to be checked and possibly changed, but I don't have a suggestion to help.

adb_graphics/default_specs.yml
```diff
@@ -740,6 +740,56 @@ icmr: # Ice Water Mixing Ratio
...
+  500ft: &icprb
+    clevs: !!python/object/apply:numpy.arange [5, 86, 10]
+    cmap: gist_ncar
+    colors: icprb_colors
+    ncl_name: ICPRB_P0_L102_{grid}
+    ticks: 10
+    transform: conversions.percent
+    unit: "%"
+    vertical_index: 0   # grib2 vertical levels are in meters,
+                        # so using index here to avoid errors from conversions
```
christinaholtNOAA, line 753: Would you mind leaving an additional comment in the code about how you chose these particular index values? Like "looked at an example file, and eyeballed which level was close to the right level" might even be enough of a clue as to how to do this in the future if something changes.

Just some kind of hint as to how these numbers were derived.

Brian-Jamison (approved):
Great job!  I have a couple of minor questions as comments in the 'Files changed'.

adb_graphics/default_specs.yml
```diff
@@ -740,6 +740,56 @@ icmr: # Ice Water Mixing Ratio
...
+    vertical_index: 29
+  max:
+    <<: *icprb
+    transform: 
+      funcs: field_column_max
+      kwargs:
+        variable: icprb
+icsev: # Icing Severity
+  500ft: &icsev
+    clevs: [0.0, 0.5, 1.0, 2.0, 3.0]
```
christinaholtNOAA, line 771: It might be good to add a quick note here that mentions the numerical values get mapped to categorical labels in maps.py and that the original data is delivered out of order, so the 0.5 value is handled in maps.py, too.

christinaholtNOAA (approved):
@cshartsough Nice work! 

I am a tiny bit concerned about the example plots included up above. The max icing severity seems to be a lower category than the 15000' one. Maybe I'm missing something there?


Just a few quick requests for more comments.

kayeekayee (approved):
Great job Craig! I am so lucky to be one of the reviewers because I learned so much by reading the code changes and all of you comments.

adb_graphics/datahandler/gribdata.py
cshartsough: yes, '1000ft' was arbitrary but has been removed.  Also, yes, axis=0 is the first array variable, in the RRFS grib2 files that's vertical.

PR #182, opened 2022-11-04
Fix for HRRR-CAR

Brian-Jamison:
Added entries in default_specs.yml to accommodate the smoke plots for HRRR-CAR.


cshartsough (approved):
good catch on this. thanks.

PR #183, opened 2022-11-07
Revisions to HWP and dust

Brian-Jamison:
Ravan had some requests in an email to me:

Can you make a few changes in the colorbars?
- Let's show HWP>10 only. So we can remove the dark blue color from the colorbar.
- For near-surface dust let's show >1 only. To use the same number of colors we can add 75 to the colorbar. 
- For ver.int.dust let's show >1 only. Let's set the maximum to 150, and change "100,200" to "75,110,150". 
Thus we don't need to show the decimal parts of the values in the colorbars for the dust plots. 

I've made these changes in this PR.

### DIFF
--- a/adb_graphics/default_specs.yml
+++ b/adb_graphics/default_specs.yml
@@ -457,7 +457,7 @@ dewp: # Dew point temperature
     wind: 10m
 dust:
   int:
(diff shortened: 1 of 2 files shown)


christinaholtNOAA (approved):

cshartsough (approved):
Thanks for all the work on this.

PR #185, opened 2022-11-29
Fix for how RRFS now handles acc WEASD.

christinaholtNOAA:
This fix should fix the issues we're seeing for RRFS_B.


cshartsough (approved):
It seemed to work well for a long time as a hot fix, so hoping nothing strange happens. :-)

Brian-Jamison (approved):
LGTM!

PR #186, opened 2022-11-29
Capability for users to create batch difference plots

christinaholtNOAA:
Users now have the option to create a "diff" plot using `create_graphics.py`. It will accept two directories with two options 
 for file name templates, and will create all the difference maps for variables specified in the image list. All other behaviors are similar to regular map plotting (forecast hours, regions, etc.).
 
 There are a few things it won't do out of the box:

- Plot differences from different models. HRRR and RRFS, for example.
- Plot graphics accumulated variables. It will plot the variables labeled as accumulated in UPP, though.
- Have a uniform color scale across variables. Each plot will assess the min/max values of the data field and use those numbers to set the value range.
- Use different color maps. That's hard coded.

Here is an example plot:
image

I'll add some additional comments to the code after the PR is opened.

### DIFF
--- a/adb_graphics/figure_builders.py
+++ b/adb_graphics/figure_builders.py
@@ -140,7 +140,7 @@ def parallel_maps(cla, fhr, ds, level, model, spec, variable, workdir,
                        '2nd set of files. Skipping.'))
                 return
 
-            field.values = field.values() - field2.values()
+            field.data = field.values() - field2.values()
 
 
         map_fields = maps.MapFields(
--- a/adb_graphics/figure_builders.py
+++ b/adb_graphics/figure_builders.py
@@ -52,9 +52,10 @@ def add_obs_panel(ax, model_name, obs_file, proj_info, short_name, tile):
     dm.draw(show=True)
 
 def parallel_maps(cla, fhr, ds, level, model, spec, variable, workdir,
-                  tile='full'):
+                  tile='full', ds2=None):
 
     # pylint: disable=too-many-arguments,too-many-locals
+    # pylint: disable=too-many-branches,too-many-statements
 
     '''
     Function that creates plan-view maps, either a single panel, or
@@ -82,8 +83,11 @@ def parallel_maps(cla, fhr, ds, level, model, spec, variable, workdir,
     last_panel = False
 
     # Declare the type of object depending on graphic type
-    map_class = maps.MultiPanelDataMap if cla.graphic_type == \
-        'enspanel' else maps.DataMap
+    map_classes = {
+        "enspanel": maps.MultiPanelDataMap,
+        "diff": maps.DiffMap,
+        }
+    map_class = map_classes.get(cla.graphic_type, maps.DataMap)
 
     for index, current_ax in enumerate(axes):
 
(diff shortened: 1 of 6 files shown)


adb_graphics/figure_builders.py
```diff
@@ -52,9 +52,10 @@ def add_obs_panel(ax, model_name, obs_file, proj_info, short_name, tile):
     dm.draw(show=True)
 
 def parallel_maps(cla, fhr, ds, level, model, spec, variable, workdir,
-                  tile='full'):
+                  tile='full', ds2=None):
 
     # pylint: disable=too-many-arguments,too-many-locals
+    # pylint: disable=too-many-branches,too-many-statements
```
christinaholtNOAA, line 58: This function could likely use some refactoring given this growing list of disabled linting concerns.

cshartsough (approved):
Looks good.  I can follow along with it, but would never have come up with it on my own.  ;-)

adb_graphics/figure_builders.py
```diff
@@ -117,6 +121,27 @@ def parallel_maps(cla, fhr, ds, level, model, spec, variable, workdir,
...
+                )
+
+            try:
+                field2.field
+            except errors.GribReadError:
+                print((f'Cannot find grib2 variable for {variable} at {level} in',
+                       '2nd set of files. Skipping.'))
+                return
+
+            field.values = field.values() - field2.values()
```
christinaholtNOAA, line 143: I actually need to fix this. We will need a setter/getter to make this work appropriately!

adb_graphics/figures/maps.py
```diff
@@ -392,7 +392,7 @@ def _draw_field(self, ax, field, func, **kwargs):
         '''
 
         x, y = self._xy_mesh(field)
-        vals = field.values()
+        vals = field.values
```
christinaholtNOAA, line 395: This will also need to change.

adb_graphics/specs.py
```diff
@@ -26,15 +26,19 @@ def __init__(self, config):
         with open(config, 'r') as cfg:
             self.yml = yaml.load(cfg, Loader=yaml.Loader)
 
-    def centered_diff(self):
+    def centered_diff(self, cmap=None, nlev=None):
```
christinaholtNOAA, line 29: I really enjoyed that we already had this option!

create_graphics.py
```diff
@@ -332,19 +336,33 @@ def parse_args():
         )
 
     # Ensemble panel-specific args
-    map_group = parser.add_argument_group('Ensemble Panel Arguments')
-    map_group.add_argument(
+    ens_group = parser.add_argument_group('Ensemble Panel Arguments')
+    ens_group.add_argument(
```
christinaholtNOAA, line 340: This was mostly unrelated, and I have no idea what kinds of issues it may have caused!

Brian-Jamison (approved):
Very nice and very comprehensive!  I'm eager to see it in action!

You mentioned some of the things it can't do, and the first was use different models.  Actually, it should be able to diff RRFS_B and HRRR, because I think they're using the same domain.

Anyway, great work!  This will be a nice addition.  If you can, I would record the instructional meeting that you're having tomorrow.

christinaholtNOAA:
@Brian-Jamison I am not totally convinced all variables will work given the difference in naming convention between the two models -- HRRR and RRFS. It might be true that for some select variables that do share the same variable names, these might work, but I don't want to claim that it certainly would.

PR #187, opened 2022-11-30
Fix for finding grid suffix appropriately.

christinaholtNOAA:
This modification was applied in the gribfile class, but needed to be here, too.


cshartsough (approved):
Thanks.

PR #188, opened 2022-11-30
Add a README for diff plots.

christinaholtNOAA:
Just a little bit of info for this tool.


cshartsough (approved):
good thinking

Brian-Jamison (approved):
Thanks, looks good!

PR #191, opened 2022-12-07
initial changes for SLW colors to match NCL version.

cshartsough:
This changes the color table for supercooled liquid water (SLW) to match the old NCL colors, which seemed to be a preference for Steve W.  It also fixes an issue with a placeholder field for SLW in default_specs which was in HRRR but not in RRFS.

passed pylint and pytest.

sample of NCL, old python, and new python colors below.
image
image
image


christinaholtNOAA (approved):
LGTM.

Brian-Jamison (approved):
Looks good, thanks!

PR #193, opened 2023-01-04
fix ncl_name for ice pellets (CICEP)

cshartsough:
Trevor Alcott notices an issue with the precip type plots (ptyp) for the sleet (ice pellets), CICEP.  It turned out that the wrong ncl_name was being used in default_specs.yml.  That's been corrected.

Passed pylint and pytest.
samples of Trevor's plot, old pygraf plot and new pygraf plot below.
image
Screen Shot 2023-01-04 at 4 15 04 PM


Brian-Jamison (approved):
Looks good!

PR #195, opened 2023-02-13
Run max

cshartsough:
This change adds two new methods, run_max and run_min, analogous to run_total, for use with run-total helicity.  Helicity runs should show the maximum/minimum over the course of the forecast run, rather than the summed amounts.

Passed pylint and pytest.

Examples below.

Before and after, 0-3 km min helicity, EastCO domain:
image
image

Before and after, 0-3 km max helicity, EastCO domain:
image
image

### DIFF
--- a/adb_graphics/datahandler/gribdata.py
+++ b/adb_graphics/datahandler/gribdata.py
@@ -671,6 +671,22 @@ class fieldData(UPPData):
 
         return vals
 
+    def run_max(self, values, **kwargs):
+
+        ''' Finds the max hourly value over all the forecast lead times available. '''
+
+        # pylint: disable=unused-argument,no-self-use
+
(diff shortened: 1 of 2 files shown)


Brian-Jamison (approved):
Looks good!  Although it's certainly believable that the lowest min value over a 12hr period would not get lower than -25, it would be good to have an example where we can see something plotted for that plot.  It might be difficult to find a good example, though.

cshartsough:
adding plots of 0-3 km minimum helicity, EastCO domain, 12-hr forecast.  I cut down the contour levels by about 100x (for this test only, not part of PR).   Before (run_total) and after (run_min):
image
image

Brian-Jamison:
Thanks for doing this test!  Looks like it's working fine.

PR #198, opened 2023-02-16
Changes to make wind contour levels above 250mb to be the same as 250mb.

Brian-Jamison:
Stan requested that wind levels above 250mb have the same contour levels as 250mb.
Tested with global data (HRRR and probably RRFS don't have these higher levels).
Passed pylint and pytest.

image
image
image
image


christinaholtNOAA (approved):
LGTM

PR #200, opened 2023-02-17
change name of surface pressure field from PRMSL to MSLET

cshartsough:
From Eric James:
"(1) There was a decision to unify our sea level pressure output with the "membrane" sea level pressure approach, which has the GRIB2 variable MSLET.  The former MAPS SLP, which was under PRMSL, is not going to be output any more.  So it would be great if we could make sure we are using the MSLET variable in any of our SLP graphics."

This change was necessary to prevent crashing the code for some precip fields which overlay MSLP.  Testing with realtime data and watching to see if it worked as a hot fix.

Passed pylint and pytest.


Brian-Jamison (approved):
LGTM!

PR #201, opened 2023-02-22
VISBSN plots

cshartsough:
From Eric:
We just started outputting an additional visibility variable which includes the effect of blowing snow on the surface visibility. It's labeled as VISBSN, but with wgrib2 it shows up like this:
44:41424492:vt=2023021621:surface:3 hour fcst:var discipline=0 master_table=2 parmcat=19 parm=35:
Would it be possible to get graphics of this added to the web? We could just put in a new row under the existing visibility, labeled something like "visibility incl. blowing snow (experimental)". Note that it will only show up for RRFS-B, not HRRR or any other models. It would be good if we could do some evaluations of the effect of blowing snow, while we still have some winter remaining.

Changes to default_specs and image_lists/rrfs_subset to add this field.  Eric has a UPP fix in to allow the plots to be generated now.  Brian is adding a place on the RRFS_B Web page for the new field.

passed pylint and pytest.

Sample vis and visbsn plots below.
image
image

### DIFF
--- a/adb_graphics/default_specs.yml
+++ b/adb_graphics/default_specs.yml
@@ -1624,7 +1624,7 @@ vil: # Vertically Integrated Liquid
     ncl_name: VAR_0_16_201_P0_L10_{grid}
     title: Vertically Integrated Liquid
 vis: # Visibility
-  sfc:
+  sfc: &vis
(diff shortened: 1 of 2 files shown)


Brian-Jamison (approved):
LGTM!

PR #202, opened 2023-02-23
update high end of Tsfc color scale in specs.py

cshartsough:
From Tanya:
Brian and Craig, will it be possible to change the color table on the temperature plots towards the right end when T > 90 F on the right plot and T>314K on the left plot. The current color table is confusing because it is impossible to say if temperature in the very southern part of Texas on the border with Mexico is 100F or 72F. Thank you!

image

I changed the colors at the top end to a brown scale.  Waiting for Tanya to make sure it satisfies the need.  Sample before/after below.

Passed pylint and pytest.

colors


Brian-Jamison (approved):
LGTM, thanks for doing this!

christinaholtNOAA:
Overcome by events.

PR #203, opened 2023-02-28
Correcting FRZR problem for RRFS.

Brian-Jamison:
This was a hot fix that was put in place because the FRZR variable needed to be renamed at later hours for RRFS.  This was tested with RRFS_A, RRFS_B, and HRRR.


cshartsough (approved):
Looks good.  I'd look more but I get PTSD looking and any accumulating variables and run total code.  ;-)

PR #206, opened 2023-03-14
Prep for Testing

brendan-ryan:
Added import sys.
Added parameter to parse_args function and argument to function call.
Wrapped command line argument checks in function check_args.
Calling check_args from main conditional.

### DIFF
--- a/create_graphics.py
+++ b/create_graphics.py
@@ -697,12 +697,11 @@ if __name__ == '__main__':
         CLARGS.images = load_images(CLARGS.images)
         CLARGS.tiles = generate_tile_list(CLARGS.tiles)
 
-    # Make sure the second data root is provided when doint diffs
+    # Make sure the second data root is provided when doing diffs
     if CLARGS.graphic_type == 'diff':
         if not CLARGS.data_root2:
-            errmsg = "Must specify a second data root for creating difference maps"
-            print(errmsg)
-            raise argparse.ArgumentError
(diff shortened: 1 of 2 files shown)


create_graphics.py
```diff
@@ -722,3 +726,7 @@ def graphics_driver(cla):
         if name not in ['specs', 'sites']:
             print(f"{name:>15s}: {val}")
     graphics_driver(CLARGS)
+
+
+if __name__ == '__main__':
+    check_args(sys.argv[1:])
```
christinaholtNOAA, line 732: Since this is the main function of create_graphics, it may be slightly easier to follow if the name of this function is `create_graphics`.

christinaholtNOAA (approved):
@brendan-ryan Thanks for this addition. It will make it much more straightforward to add integration tests for this tool.

create_graphics.py
brendan-ryan: That makes sense. Is this something we should change now or in a subsequent PR?

christinaholtNOAA: Do you mind pushing it to this one?

brendan-ryan: Just pushed a new commit. Github should automatically update the PR, correct?

christinaholtNOAA: That's right. :)

cshartsough (approved):
Nice work.  Thanks for the effort, and to Christina for her comments/suggestions.

PR #208, opened 2023-03-22
Global acc var fix

Brian-Jamison:
Global models were recently extended to 336h, which revealed a change in the accumulated precip variable that happens at 246h, after which there are 12h accumulations.  This affects the renaming procedure, and crashes on the new variable.
This fix simply drops those new 12h accumulations.


cshartsough (approved):
We're getting used to these kinds of lists.  Hoping some day the UPP will sort all this out, but looks like this will do until then.  Thanks.

PR #209, opened 2023-03-24
Updated colorbars and full vertically integrated dusts for dust plots

kayeekayee:
1) Changed the range of the colorbars.
near_sfc_fine_dust
near_sfc_coarse_dust
2) Combined the vertically integrated dusts (fine and coarse) into one plot.
full_int_dust

### DIFF
--- a/adb_graphics/default_specs.yml
+++ b/adb_graphics/default_specs.yml
@@ -410,17 +410,8 @@ clwmr: # Cloud water Mixing Ratio
       nat: CLWMR_P0_L105_{grid}
       prs: CLWMR_P0_L100_{grid}
 coarsedust:
-  int: &coldust
-    clevs: [1, 2, 3, 4, 6, 8, 10, 15, 20, 30, 50, 75, 110, 150]
-    colors: smoke_colors
-    ncl_name: COLMD_P48_L200_{grid}_A62001_1
-    plot_airports: False
-    ticks: 0
-    title: Vertically Integrated Coarse Dust
-    transform: conversions.to_micro
-    unit: $mg/m^2$
-  sfc: &sfcdust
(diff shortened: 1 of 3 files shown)


cshartsough (approved):
These changes look good.  If you think you'll ever want plots of integrated coarse dust only, you could leave the parameters in the default_specs file (they are still there for integrated fine dust).

cshartsough (approved):
These changes look good.  If you think you'll ever want plots of integrated coarse dust only, you could leave the parameters in the default_specs file (they are still there for integrated fine dust).

Brian-Jamison (approved):
Thanks, Ka Yee!  This looks great!

cshartsough:
It looks like the automated tests failed due to the new summed field having "_" in the name, which our pygraf interprets in some cases to look for special hours, etc.  If you change full_int_dust to something like fullintdust, etc., it should work.  Sorry I didn't think of that when I gave you the example.

adb_graphics/default_specs.yml
```diff
@@ -480,13 +471,38 @@ dewp: # Dew point temperature
     wind: 10m
 finedust:
   int:
-    <<: *coldust
+    clevs: [1, 4, 7, 11, 15, 20, 25, 30, 40, 50, 75, 150, 250, 500]
```
christinaholtNOAA, line 474: You could add a PyYAML anchor here and reference it later in full_int_dust to reduce duplication, like was done for coldust before this change.

christinaholtNOAA (commented):
@kayeekayee This looks good to me. Just one minor suggestion below.

adb_graphics/default_specs.yml
kayeekayee: Thank you Christina!

kayeekayee:
Thank you so much for reviewing this PR. Here is some updated plots for the new fire yesterday.
Vertical_intergated_dusts
fine_dust
coarse_dust

PR #210, opened 2023-04-12
Changed units for smoke emissions and changes to allow RRFS_A to run.

Brian-Jamison:
Smoke emissions units should be μg m-2 s-1.  This change is made.
Also, code changes to allow RRFS_A to run are implemented.  This has been tested through 60 hrs for RRFS_A, 24 hrs for RRFS_B, and 24 hrs for a global model.

Passes pylint and pytest.

### DIFF
--- a/adb_graphics/datahandler/gribfile.py
+++ b/adb_graphics/datahandler/gribfile.py
@@ -127,6 +127,16 @@ class GribFiles():
                         suffix == f'{suf}1h':
                         contains_suffix.append(suf)
 
+                    # RRFS_A has fields that have the suffix 'acc0h' but we don't
+                    # want those. Drop them if they come up.
+                    bad_0h_vars = ['APCP_P8_L1_GLL0_acc0h', \
+                        'FROZR_P8_L1_GLC0_acc0h', 'FRZR_P8_L1_GLC0_acc0h', \
+                        'CDLYR_P8_L200_GLC0_avg0h', 'TCDC_P8_L200_GLC0_avg0h', \
+                        'APCP_P8_L1_GLC0_acc0h', 'APCP_P8_L1_GST0_acc0h']
+                    if fhr != 0 and var in bad_0h_vars:
+                        print(f'dropping {var}')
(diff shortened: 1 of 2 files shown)


adb_graphics/datahandler/gribfile.py
```diff
@@ -127,6 +127,16 @@ def free_fcst_names(self, ds, fcst_type):
                         suffix == f'{suf}1h':
                         contains_suffix.append(suf)
 
+                    # RRFS_A has fields that have the suffix 'acc0h' but we don't
```
cshartsough, line 130: If these fields are based on my (CSH) suggestions, there may be some missing or some that don't need to be in the list.  We'll need to monitor the list as more fields are added to the grib files and requested for graphics.

adb_graphics/default_specs.yml
```diff
@@ -922,7 +922,7 @@ emissions: # Smoke emissions
     ticks: 0
     title: Smoke Emissions
     transform: conversions.to_micrograms_per_m3
-    unit: $\mu g/m^3$
```
cshartsough, line 925: Assuming we don't need to change the conversion calculation, that this is just the preferred convention for the units?  Do we need to change the name of the transform?

cshartsough (approved):
Looks pretty good.  Two features in this, but both small and related, so should be okay.  I added a couple minor comments.  Thanks for the hard work.

adb_graphics/datahandler/gribfile.py
Brian-Jamison: Yes, we should monitor the list.  Hopefully, it won't need to be changed for a long time.

adb_graphics/default_specs.yml
Brian-Jamison: I'm pretty sure we don't need to change the conversion.  The original units for this field are kg m-2 s-1, so we are really only converting the kg to micrograms.  I don't think it's necessary to change the name of the transform, even though it's a bit of a misnomer, in this case.

PR #221, opened 2023-08-03
Extend and add colors for max updraft helicity, hourly and run total

cshartsough:
This change extends the upper limit for maximum updraft helicity, both hourly and run total.  The change was requested by Jack Settelmaier through Jacob Carley and Steve Weygandt.

Change includes using "join_ranges" method to keep the number of new colors limited (from 12 to 16 levels) while roughly doubling the upper limit. Fields changed are 0-3km and 2-5km max UH (also changed 0-2km and 1-6km, but those aren't currently plotted.)

passed pylint.

Samples below (old and new).
image
image

image
image


Brian-Jamison (approved):
LGTM!  Thanks for doing this!

PR #223, opened 2023-08-28
use temperature colors for dewp point temp

cshartsough:
this change was requested by Stan, based on discussions with other end users.  There was consensus that dew point fields should use the same color table as temperatures, to make comparisons easier.

Change was made to `default_specs.yml`.  The `dewp_colors` table was removed from the specs.py file as well.

before/after examples below.

passed pylint.

image
image


jaymes-kenyon (approved):

Brian-Jamison (approved):
Great job!  Thanks!

PR #227, opened 2023-10-05
bug fixes to run graphics with hafs-ar (arfs) on hera

SamuelTrahanNOAA:
I'm able to run the graphics on HAFS with these changes.


adb_graphics/datahandler/gribfile.py
```diff
@@ -218,7 +218,7 @@ def _load(self, filenames=None):
                     )
 
                 renaming = self.free_fcst_names(dataset, fcst_type)
-                if renaming and self.model not in ['hrrre', 'rrfse']:
+                if False: # renaming and self.model not in ['hrrre', 'rrfse']:
```
Brian-Jamison, line 221: This statement is new to me (but I'm fairly new to python).  From what I read on StackOverflow, "if False:" prevents the code that follows from executing.  This may be ok for HAFS, but I think would break runs for other models which need the renaming.

Also, I'm not sure what happened to the create_graphics.py script.  It says, "Empty file."  Did it get deleted?

SamuelTrahanNOAA: I was unable to find another workaround for the exception raised in renaming.

I'm not sure what "Empty file" means there, either. The file shouldn't be empty. If it was deleted, the github diffs would say that and the red -2 at the top of the page would be a much bigger number.

SamuelTrahanNOAA: I just checked, and the file certainly isn't empty. The "Empty file" message is a glitch in the github website.

Brian-Jamison: I think the way I would do it is to have a separate yml file for the ARFS image list.  The model name is retrieved from there.  It could be "arfs" in this case.  Then add it as another exception for self.model, like this:
  if renaming and self.model not in ['arfs', 'hrrre', 'rrfse']:
Does this sound like it would work?

SamuelTrahanNOAA: I have implemented a proper solution. When free_fcst_names produces duplicate variable names, they'll be flagged and skipped. You get a message like this instead of a fatal exception from xarray:

```
Loading grib2 file: free_fcst, /scratch2/BMC/wrfruc/Samuel.Trahan/scrub/UPPSteveCase/com/2023022400/00L/00l.2023022400.hfsa.storm.atm.f012.grb2
RENAMING VARIABLES FOR MODEL "global":
          APCP_P8_L1_GLL0_acc12h  -> APCP_P8_L1_GLL0_acc
  ERROR: Renaming would produce a duplicate:          APCP_P8_L1_GLL0_acc3h  -> APCP_P8_L1_GLL0_acc
        TSNOWP_P8_L1_GLL0_acc12h  -> TSNOWP_P8_L1_GLL0_acc
  ERROR: Renaming would produce a duplicate:        TSNOWP_P8_L1_GLL0_acc3h  -> TSNOWP_P8_L1_GLL0_acc
          FRZR_P8_L1_GLL0_acc12h  -> FRZR_P8_L1_GLL0_acc
  ERROR: Renaming would produce a duplicate:          FRZR_P8_L1_GLL0_acc3h  -> FRZR_P8_L1_GLL0_acc
         FROZR_P8_L1_GLL0_acc12h  -> FROZR_P8_L1_GLL0_acc
  ERROR: Renaming would produce a duplicate:         FROZR_P8_L1_GLL0_acc3h  -> FROZR_P8_L1_GLL0_acc
         TCDC_P8_L200_GLL0_avg3h  -> TCDC_P8_L200_GLL0_avg
        CDLYR_P8_L200_GLL0_avg3h  -> CDLYR_P8_L200_GLL0_avg
Internal error: free_fcst_names produced duplicate variable names. To avoid a fatal exception from xarray, some renaming has been skipped.
```

christinaholtNOAA (approved):
LGTM.

We may need a bit of refactor to get it into main, but this is fine in arfs_dev.

SamuelTrahanNOAA:
Could someone please merge this?

It is the last submodule pointing to my fork in the Atmospheric River fork of HAFS.

PR #230, opened 2023-10-25
Global chem products

Brian-Jamison:
Changes to plot global chem products.  A couple of examples:

image
image


cshartsough (approved):
This looks good. Makes sure your column total default_specs are in the right place alphabetically, otherwise it looks ready to go.

PR #232, opened 2024-01-05
Fix rrfs a issues

Brian-Jamison:
These changes correct the problems with the numerous black boxes from contour labels.  Below is an example:

image

The number of contours has been changed as well, from every 2mb to every 8mb, reducing the cluttered contour lines.

Accumulated snow and precip plots are working:

image
image

Additionally, Craig's new color table for skin and 2m temperature has been implemented.

### DIFF
--- a/adb_graphics/default_specs.yml
+++ b/adb_graphics/default_specs.yml
@@ -1577,8 +1577,8 @@ temp: # Temperature
     wind: False
   2m: &sfc_temp
     annotate: True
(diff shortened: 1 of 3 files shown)


cshartsough (approved):
looks good.  You have the old clab routine commented out, it could be removed since we can find it in the old PRs if we ever need it.  :-)

jaymes-kenyon (approved):

PR #233, opened 2024-01-13
Fix RRFS_A issues part 2

Brian-Jamison:
Changed color table back to 10 deg contours.  Added ground heat flux plot.  Example below.

image


cshartsough (approved):
good work.  thank for dealing with these issues.

PR #234, opened 2024-01-26
Fix RRFS A issues part 3

Brian-Jamison:
Changes to prevent 2m temp annotation in large domains like RRFS NA and RAP full domains.  Also added acsnod plot for RAP and corrected the 1h snowfall for RRFS.  MSLP contours are also reduced for the large domains.


cshartsough (approved):
All makes sense to me.  Thanks for the hard work.

PR #235, opened 2024-02-26
WFIP3 Domains for HRRR

Brian-Jamison:
These are changes necessary for comparing HRRR output to WFIP3, both the full and nested domains.  Also, 160m and 320m AGL wind plots have been added.

image
image
image

Sorry, the HRRR image does not have the same time as the others.


cshartsough (approved):
Looks good.  It looks like the transform segments for 160m/320m winds in default_specs are the same as the &ua_wpeed values?  If so they could be removed.

Thanks for making these changes.

Brian-Jamison:
Thanks, Craig!  I removed that for 80m wind as well.

PR #236, opened 2024-02-28
Add a model-name dependence for some cloud fields

jaymes-kenyon:
This update uses the model name to correctly assign handles for the ceiling (exp-2) and cloud-base height fields.

### DIFF
--- a/adb_graphics/default_specs.yml
+++ b/adb_graphics/default_specs.yml
@@ -413,7 +413,12 @@ ceilexp: # Ceiling - experimental
 ceilexp2: # Ceiling - experimental no.2
   ua:
     <<: *ceil
-    ncl_name: CEIL_P0_L2_{grid}
+    ncl_name: 
+      hrrr: HGT_P0_L2_{grid}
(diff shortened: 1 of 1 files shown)


cshartsough (approved):
Looks good.  Are you anticipating more models having cloud-base graphics, or just using the model name there for completeness/symmetry?

jaymes-kenyon:
Hi @cshartsough — The main rationale for this PR is the ceiling-exp2 field.  A previous PR changed the handle for this field to work with RRFS output, but the same PR inadvertently caused HRRR graphics to stop for this field.  With this update, hopefully we will see the ceiling-exp2 plots from both the RRFS and HRRR (each using different handles for the same field).

cshartsough:
That's good, I think it should do the trick nicely.

jaymes-kenyon:
I am going to let this run in a RRFS retro overnight, just as a quick test.  I will follow-up once I have some successful plots.

jaymes-kenyon:
@cshartsough @Brian-Jamison — My tests have worked, so I think this PR is ready for merging.

PR #239, opened 2024-04-11
Skip Analysis

Brian-Jamison:
This is a small change to ignore the precip type plot at analysis.  It is needed because the wgrib2 command (that converts the wind components to Earth-relative) removes the categorical precip fields at analysis, presumably because they are all zeroes.  This causes pygraf to crash.  This change corrects that.  One caveat is that the --all_leads flag needs to be removed, since ignoring the precip type fields causes a different error at forecast hour 1.


Brian-Jamison:
Sorry, it occurred to me that this is not model-specific, and may cause problems with models other than HRRR_HI.
I'm putting in a hotfix for HRRR_HI, and closing this PR.

PR #240, opened 2024-05-16
Adding LWP and IWP to default_spec.yml

XiaSun-Atmos:
Adding LWP and IWP to default_spec.yml, with unit as g/m2. 
This is tested on WFIP3 (WRF runs) UPP output. It should also work for HRRR.
Example plots as below:
lwp_WFIP3-d01_sfc_f001
iwp_WFIP3-d01_sfc_f001

### DIFF
--- a/adb_graphics/conversions.py
+++ b/adb_graphics/conversions.py
@@ -30,7 +30,7 @@ def kg_to_g(field, **kwargs):
     ''' Conversion from kg to g '''
 
     return field * 1000.
-    
+
 def magnitude(a, b, **kwargs):
 
     ''' Return the magnitude of vector components '''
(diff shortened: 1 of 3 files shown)


cshartsough (approved):
Looks to be set up well.  The high end of the range seems to be saturated, but I wouldn't be too worried about extending it unless you think users will notice and ask for it.

cshartsough (commented):
The lint test failed due to extra whitespace at the end of the new kg-to-g conversion.  I hate those, but a quick fix, at least.

XiaSun-Atmos:
Thanks @cshartsough ! 
The high end of the range does look saturated. The current colorbar range is per the users' (the local wfip team) request, so it should be good for a while;
Removed the extra whitespace at the end of the new kg-to-g conversion;
The last commit is to add lwp and iwp to wfip image lists.

PR #241, opened 2024-07-08
initial changes to plot HFIP products.

cshartsough:
HFIP runs have started and currently have a small number of output plots.  These small changes allow the "HFIP" domain to be plotted, and set the HFIP dimensions to be treated as 1D, as global plots.

Sample plots are below; additional domain and plots may need some extra changes. e.g.,
1. upper level temperatures are currently plotting as 0.5 deg K everywhere.  seems to be in the grib2 data.
2. we don't have ability to plot specific humidity yet.
3. plotting the 'full' grid plots the grids upside down.

Passed pylint.

image
image
image

### DIFF
--- a/adb_graphics/datahandler/gribdata.py
+++ b/adb_graphics/datahandler/gribdata.py
@@ -551,7 +551,7 @@ class fieldData(UPPData):
         '''
 
         lat, lon = self.latlons()
-        if self.model in ['global', 'obs']:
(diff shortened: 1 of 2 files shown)


Brian-Jamison (approved):
Looks good!  Seems like a weird domain, but it covers the area of interest!

I'm not sure how hard it would be to flip the domain for the full domain, but if it's a separate file, it probably should be done in the post-processing.

Thanks for taking this on!

PR #242, opened 2024-07-22
Changes for RRFS_A adding emissions and lightning, and changing to u…

Brian-Jamison:
…se the post-processed value for Hourly Wildfire Potential (firewx) rather than computing it in pygraf.


cshartsough (approved):
Glad to have the calculations outside of the graphics now.  Thanks.

PR #244, opened 2024-07-31
changes to flux color scales and table, 3 separate color schemes

cshartsough:
Stan B., Dave T., and Jeff D. has recommendations for new color schemes for latent/sensible/ground heat flux plots.  They wanted different scales/colors for each, so I've separated the heat_flux_colors in specs.py into three different sets, each patterned after sample supplied by Jeff.D.  I left the original in specs.py for now.

Sample plots below.

Passed pylint.

image
image
image


Brian-Jamison (approved):
LGTM!  Thanks for doing these.  Hopefully it will make everyone happy!

PR #245, opened 2024-08-06
RRFS lightning with precip

Brian-Jamison:
Adding a new plot, which is an amalgamation of a new field (lightning) and 1h total precip.

Here's an example:

image


cshartsough (approved):
I don't see any issues, I assume it will get some wide usage.

PR #246, opened 2024-09-09
Changes to RRFS lwtp plot.

Brian-Jamison:
These are some small changes to modify the lwtp (lightning with total precip) plot.  Now we only plot above 0.25 precip to more easily see lightning away from high precip.

image


cshartsough (approved):
Looks like a good idea to reduce the precip plot levels.  Should be a well-used product.

PR #247, opened 2024-11-14
fix for special accumulation case where MASSDEN 0h field ends in '_1'

cshartsough:
The MASSDEN field (measuring dust) has two variables, one for coarse and one for fine. For the initial hour, one field ends with '_avg' and one with '_avg_1'. For later hours they end in '_avg1h' and '_avg1h_1'.

The pygraf code recognizes the '_avg' and renames the field as needed, but doesn't recognize the '_avg_1' field as needing to be renamed. The error causes the graphics code to completely crash on hour=001.

This change treats the initlal-hour '_avg_1' field separately from the other code logic, and does the necessary renaming.  

Tested in the RRFS_B realtime runs for twelve hours, seems to work as desired.

Passed pylint. Will do pytest on PR submission.


christinaholtNOAA (approved):
LGTM.

PR #251, opened 2025-02-22
new_ua_temp_colormaps

Brian-Jamison:
These are some requested changes in the 925mb, 850mb, 700mb, and 500mb temperature contours, to clearly delineate the 0 deg line and provide more clarity in the -5 to +5 C range.  The 500mb scale was changed to match the scales at the other levels.

image
image

### DIFF
--- a/adb_graphics/default_specs.yml
+++ b/adb_graphics/default_specs.yml
@@ -1674,7 +1674,7 @@ temp: # Temperature
     unit: F
     wind: 10m
   500mb: &ua_temp
-    clevs: !!python/object/apply:numpy.arange [-70, 10, 2.5]
+    clevs: !!python/object/apply:numpy.arange [-40, 40, 2.5]
     cmap: jet
     colors: t_colors
     contours:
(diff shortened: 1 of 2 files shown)


cshartsough (commented):
I see the pytest failed but it wasn't clear right offhand what the issue is.  I can help look at it more if you don't see the problem either.  Otherwise it seems to fit the logic pretty well, so should be able to approve after this fix.

Brian-Jamison:
> I see the pytest failed but it wasn't clear right offhand what the issue is. I can help look at it more if you don't see the problem either. Otherwise it seems to fit the logic pretty well, so should be able to approve after this fix.

I'm not quite sure what the problem is, either.  The range(17, 1, -2) gives 17, 15, 13, 11, 9, 7, 5, 3 (8 values) and that's done twice, so 16 values below zero, and 16 other values in the "ncar" part of the table, above zero.  That matches the number of colors in the current plot, so I don't know exactly what's wrong.

cshartsough:
The test thinks that clevs=15 and t_colors =32, but I don't see why that's the case.  I see you removed the unnecessary clevs lines once &ua_temp  changed, maybe that's throwing things off?  I can try to test it on Monday.

cshartsough (approved):
Seems like a good solution for the last issue, let t_colors stay as is and use the new color table you had set up as a separate spec.

Good job.

PR #252, opened 2025-02-22
change skin temperature title from "sfc" to "Skin" Temperature

cshartsough:
A simple one-line code addition in default_spec.yml, to change the title for sfc temperature to "Skin Temperature".

This was requested by Stan Benjamin, who noted that using the name sfc temperature can be ambiguous.  The title was a default value since none was given for that field. Added a title line in default_specs.yml.

Before and after images attached. Stan approved the change.

Passed pylint.

image
image


Brian-Jamison (approved):
LGTM!  Thanks for doing this!

PR #255, opened 2025-03-20
global_chem_changes

Brian-Jamison:
Contour level changes in chem variables.  Added dlwrf and dswrf (though not currently in grib2 files), added new products lhtflavg and shtflavg (to distinguish them from the instantaneous lhtfl and shtfl).  global_chem.yml has added ctop, dewp, flru, pres (msl), ptyp, and others.  Only a few can currently plot but will leave the others in the presumption that these variables are intended to be added.

New product examples:

image

image


cshartsough (approved):
looks thorough.  thanks for taking this on.

PR #257, opened 2025-04-11
Adding "regional_mpas" as a model configuration

jaymes-kenyon:
This update adds "regional_mpas" as a model configuration.  Currently (i.e., prior to this PR), GSL's regional MPAS configurations invoke pygraf using "hrrr" as the model configuration.  By creating this "regional_mpas" functionality, pygraf can now be freely adapted to support emerging MPAS plotting compatibility, but without risking longstanding HRRR compatibility. 

This "regional_mpas" configuration has an associated YAML file, and has several unique ncl_name handles (i.e., pointers to GRIB2 fields).  When/if the regional MPAS configurations are given a formal name for operational implementation (e.g., RRFSv2 or HRRRv5), the code instances of "regional_mpas" can be renamed, as needed.

Testing has been successful on Jet using recent GRIB2 files from the "MPAS_physics_dev1" system.

### DIFF
--- a/adb_graphics/datahandler/gribdata.py
+++ b/adb_graphics/datahandler/gribdata.py
@@ -341,7 +341,7 @@ class UPPData(specs.VarSpec):
         entire atmosphere in HRRR, while 200 is used in RRFS. '''
 
         if self.filetype == 'prs':
-            if self.model == 'rrfs':
+            if self.model == 'rrfs' or self.model == 'regional_mpas':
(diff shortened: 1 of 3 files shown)


cshartsough (approved):
Looks good.  There may a couple hidden code bits that will need updating, not sure, but if so they should be easy to correct once this outline is in place.  Thanks for taking this on.

PR #258, opened 2025-04-16
Revising plots of vertical shear; enabling plots of cloud base and max updraft for regional MPAS

jaymes-kenyon:
This PR accomplishes the following changes:

1) Plots of 0-1 and 0–6-km shear are revised for all model configurations (e.g., HRRR, RRFS, MPAS).  The proposed updates include clarified title bars / descriptions, re-expressed units (kts), and modified color bars.  The intent is to improve the usability of these shear plots, especially for characterizing convective environments.  Examples of the existing plots (left) and revised plots (right) are shown below:

Regarding these shear plots, I attempted to overlay the shear vectors (represented as barbs), but I could not get pygraf to produce the overlay (the attempted code crashed).  For now, we continue to plot only the shear _magnitude_.  Advice is welcome!

2) For regional MPAS applications, plots of cloud-base height and hourly-maximum updraft velocity are enabled.

Testing has been successful on Jet using recent GRIB2 output from **MPAS_physics_dev1**.  Copying @daviddowellNOAA for awareness.

### DIFF
--- a/adb_graphics/default_specs.yml
+++ b/adb_graphics/default_specs.yml
@@ -417,6 +417,10 @@ ceilexp2: # Ceiling - experimental no.2
   ua:
     <<: *ceil
     ncl_name: 
+      # Note that the "HGT_P0_L2_{grid}" ncl_names in RAPv5/HRRRv4 (below)
+      # seemingly correspond to cloud-base height.  This is intentional; i.e.,
+      # the exp2 ceiling field was indeed implemented into operations by being 
+      # labeled as the cloud-base height.
       hrrr: HGT_P0_L2_{grid}
       hrrrhi: HGT_P0_L2_{grid}
       hrrrcar: HGT_P0_L2_{grid}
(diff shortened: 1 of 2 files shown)


adb_graphics/default_specs.yml
```diff
@@ -2163,7 +2171,9 @@ wspeed: # Wind Speed
     clevs: [0.5, 1, 1.5, 2, 2.5, 5, 7.5, 10, 12.5, 15, 17.5, 20, 22.5, 25, 30, 35, 40]
     cmap: jet
     colors: mup_colors
-    ncl_name: MAXUVV_P8_2L108_{grid}_max1h
+    ncl_name:
```
jaymes-kenyon, line 2174: Hi @Brian-Jamison — Yeah, the code below seems to work; here's a sample plot.
Screenshot 2025-04-21 at 3 08 50 PM

I decided to try this approach when I noticed a similar construction in lines 1033–1034 (for cloud-ice mixing ratio).  I am not certain, but it seems to mean "try this handle first, and if it does not work, try the next handle".  

Let me know if you foresee issues.  If HRRR plots of this field fail as a result of deploying these changes, I can revisit the lines.

Brian-Jamison: Thanks for taking a look at that, Jaymes.  If it works the way that you indicate, then it could be used to streamline some of the other code in default_specs.yml.  But that's for another time.  This looks good, as long as you're ok with the different contour levels between the 0-1 km and 0-6 km.

Brian-Jamison (approved):
LGTM!

PR #259, opened 2025-05-09
Adding plots of 1-h hail accumulation; updating plots of hail diameter and VIL

jaymes-kenyon:
acc

size

### DIFF
--- a/adb_graphics/default_specs.yml
+++ b/adb_graphics/default_specs.yml
@@ -867,9 +867,19 @@ gust:
     title: 10m Wind Gust Potential
     transform: conversions.ms_to_kt
     unit: kt
+1hachail: # 1 hr Accumulated Hail
+  sfc:
+    clevs: [0.1, 0.5, 1, 2, 3, 4, 6, 8, 10]
+    cmap: gist_ncar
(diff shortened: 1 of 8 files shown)


cshartsough (approved):
Looks good.  Two small comments:
1. If hailcast now has just one ncl_name, it can be
    "    ncl_name: HAIL_P8_L1_GLC0_max1h"
      instead of 
      "    ncl_name:
                  rrfs: HAIL_P8_L1_GLC0_max1h"
 2. Can you add a short summary to the GitHub conversation section above, on which models you're removing the Hailcast/VIL from (mostly HRRR version?) and what's added, so it's clear for future reference?

cshartsough:
Sorry for the confusion on my review, your commit explanation doesn't fully show up until I logged into GutHub, so that part is fine, just the one comment on ncl_name instead.

jaymes-kenyon:
@cshartsough — Thanks!  Good idea on the ncl_name used to extract HAILCAST; I just added that commit.

@Brian-Jamison — If it's feasible, we probably want to add a new "row" on the MPAS graphics websites for the new 1-h hail accumulation plots; see item (1) above.

PR #260, opened 2025-05-19
GFS_flux_products

Brian-Jamison:
Changes to plot 6h average flux products for GFS.  They use the same color table and contour intervals as the instantaneous plots, but are named differently and have titles that indicate that they are 6h averages.  Also a small adjustment to the title vertical position to avoid overrunning with other header lines.

image
image

These example images have different sizes, and the length of the title may somehow affect the overall image size.  I'll run some tests to see if this is the case.

### DIFF
--- a/adb_graphics/default_specs.yml
+++ b/adb_graphics/default_specs.yml
@@ -653,6 +653,16 @@ dlwrf: # Downward Longwave Radiation Flux
     ticks: 20
     title: Incoming Longwave Radiation Flux, TOA
     unit: W/m$^{2}$
+dlwrfavg: # Average Downward Longwave Radiation Flux
+  sfc:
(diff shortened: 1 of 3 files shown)


cshartsough (approved):
Looks nice.  To be clear, these are fluxes averaged over 6 hours, rather than hourly averages put out every six hours?  Just matters with the titles, the rest is good either way.

Thanks for taking this on.

Brian-Jamison:
Thanks, Craig!  Yes, I believe these are 6h averages.  The ncl_filedump utility shows (for these variables in  /public/data/grids/gfs/0p25deg/grib2/2513900000024): 

         type_of_statistical_processing :       Average
         statistical_process_duration : 6 hours (ending at forecast time)

PR #261, opened 2025-08-12
Enhance county lines

Brian-Jamison:
To increase the visibility of the county lines, they have been changed from gray to black.


cshartsough (approved):
Seems like this was a hotfix that never got formally added.  Thanks for taking care of it.

PR #262, opened 2025-09-09
Firewx compute sep2025

cshartsough:
Jun 2024, we implemented a change to pygraf to stop computing the RRFS Hourly Wildfire Potential and instead plot the field that was directly available in the grib2 files.  Unfortunately, HRRR runs didn't have this new field in their grib2 data, so the HWP for HRRR stopped working.

This change adds a new option in the default_specs.yml file to allow for either method to be used.

firewx: use the grib2 field without any transform
firewx_pygraf:  compute the field using the input vegetation field and other variables as needed for the calculation

Changes the default_specs.yml file and three hrrr-based image lists.

Passed pylint.  Pytest will be done by GitHub.


Brian-Jamison (approved):
Thanks for doing this Craig!  You corrected the underscore before I had a chance to mention it.  Don't forget to change the name in the image_lists yml files, and please include it in hrrrhi_subset.yml (HRRR HI has the VGTYP variable).  We'll change the web page plot_types.txt files once it is in place.  Great job!

cshartsough:
Thanks for the reminders.  redid the image lists names and added it to the HRRRHI images.

Will merge assuming the pytests complete okay.

PR #264, opened 2026-04-02
Fix HRRR maps on main.

christinaholtNOAA:
The HRRR image list is attempting to plot `ceilexp`, which appears to only be set up for regional_mpas. More work will be needed if this variable should actually be plotted for HRRR, too.


Brian-Jamison (approved):
LGTM!

